PR #3, opened 2018-11-27
Hero global routing

pravinblaze:
### Implementation of hero global routing api

The route is searched using Dijkstra's algorithm using a heap implementation of priority queue for fast and efficient routing.
The **GlobalRoutePlanner** class is placed in _navigation/pathrouting/global_route_planner.py_
Unit tests for the class can be found in _navigation/pathrouting/test_global_route_planner.py_

For a visual demonstration, check out this [jupyter notebook.](https://github.com/pravinblaze/carla-code-pad/blob/master/Scenario_routing_dev_pad.ipynb).

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/3)

### DIFF
--- a/navigation/pathrouting/global_route_planner.py
+++ b/navigation/pathrouting/global_route_planner.py
@@ -1,13 +1,14 @@
 # This work is licensed under the terms of the MIT license.
 # For a copy, see <https://opensource.org/licenses/MIT>.
 
-# This file contains class for global route planning
+# This file contains the class for global route planning
 
 """
 This module provides GlobalRoutePlanner implementation.
 """
 
 import math
+
 
 class GlobalRoutePlanner(object):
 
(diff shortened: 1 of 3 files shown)


felipecode:
One thing. 
It seems that you are considering all curves as intersections.
This is probably ok but conceptually you dont need to tell the agent
to turn LEFT when LEFT is the only option it has. Technically the 
four courners around Town01 are not intersections, they are just curves.

Another thing. The library you are using seems quite powerful. I wonder
why are you using Djisktra instead of A*. A* with a not overestimated heuristic
also gives you the shortest path and it is faster, since it can easily eliminate
paths that are clearly bad when following the heuristics.

I will keep reviewing. Thanks for the amazing work.

fabianoboril:
New PR: https://github.com/carla-simulator/carla/pull/1048
I will close this PR.

PR #5, opened 2018-12-05
Implementation for Traffic Scenario 10

pravinblaze:
* Added new atomic behaviour SyncArrival
* New scenario NoSignalJunctionCrossing in Scenario.no_signal_junction_crossing
* Modified scenario_runner.py to run new scenario (may be we should make this completely parametric)

Run _**python scenario_runner.py --scenario NoSignalJunctionCrossing**_
with Town03. Scenario takes place at the junction (170, 60).

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/5)


fabianoboril:
Hi Praveen,
nice work :+1: I agree with your point regarding the scenario_runner.py. I have already an idea in my mind, but need to try it out first. For now, I think we can keep it as is.

fabianoboril:
When testing your scenario, I had to problems (see my code review). Besides, a good AD stack will never cross that intersection, as the other car does not clear the crossing (it will always stay within the intersection). I think we have to add some behavior to at least vanish after some time...

fabianoboril:
One idee, how we can get this thing closed soon:
Why not letting the other car drive straight through the intersection, and the ego car coming from right making a right turn... (then the ego may go first). Or you want that the ego car makes a left turn, then it has to wait for the other car to pass. In this case, the other car always goes straight (easy to implement), and the ego-client has to make a turn (which should be doable with your routing api).

pravinblaze:
Nestor just made a new build with Xavier's changes to Town03. I'm working on it, will push the new changes by early tomorrow.

fabianoboril:
Please update Docs/list_of_supported_scenarios.md!

PR #6, opened 2018-12-07
Implementation of Traffic Scenario 03

mohit-awana:
1. Added New scenarios StationaryObjectCrash and DynamicObjectCrash in Scenario object_crash_vehicle.py

2. Modified scenario_runner.py to run new scenario.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/6)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "python.pythonPath": "/home/mohit/anaconda3/envs/scenario/bin/python"
+}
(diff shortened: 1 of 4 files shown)


fabianoboril:
I haven't tested the scenario, yet.

fabianoboril:
The scenario does still not work properly. I let the ego vehicle move with ~30 km/h. The cyclist just crosses the street. I think the cyclist has to stop. Otherwise, the ego vehicle may just avoid a collision by "luck".

fabianoboril:
Only 2 minor things (naming and the driven_distance_check) are remaining from my side. The rest is fine.

PR #7, opened 2018-12-07
Implementation of Traffic Scenario 01

SubhashiniPerumal:
* Added new scenario in control_loss.py
* Added new atomic SteerVehicle
* Added new criteria ReachedLocationTest
* Added new scenario to list in scenario_runner.py

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/7)

### DIFF
--- a/Docs/getting_started.md
+++ b/Docs/getting_started.md
@@ -25,9 +25,9 @@ pip3 install --user py_trees # For Python 3.x
 
 ## Running the follow vehicle example
 First of all, you need to get latest master branch from CARLA. Then you have to
-install the PythonAPI:
+include CARLA Python API to the Python path:
 ```
-easy_install ${CARLA_ROOT}/PythonAPI/carla-<VERSION>.egg
+export PYTHONPATH=$PYTHONPATH:${CARLA_ROOT}/PythonAPI/carla-<VERSION>.egg
 ```
 
 Now, you can start the CARLA server with Town01 from ${CARLA_ROOT}
--- a/Docs/list_of_scenarios.md
+++ b/Docs/list_of_scenarios.md
@@ -63,3 +63,8 @@ through a junction without signal.
 The ego vehicle is passing through a junction without traffic lights
 And encounters another vehicle passing across the junction. The ego vehicle has
 to avoid collision and navigate accross the junction to succeed.
+
+### ControlLoss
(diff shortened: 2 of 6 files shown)


fabianoboril:
I am not sure, if this scenario is what we really want to have.
The crash scenario describes a situation, where a vehicle lost control, due to road conditions, driving without attention, etc. Hence, I think the scenario should test, if a vehicle can drive properly from A to B (going through some narrow curves, etc.). This is also what is stated by the NHTSA crash report:
https://www.nhtsa.gov/sites/nhtsa.dot.gov/files/pre-crash_scenario_typology-final_pdf_version_5-2-07.pdf (page 29)

> Typical Scenario: Vehicle is turning left or right at an
> intersection-related area, in daylight, under clear weather
> conditions, with a posted speed limit of 45 mph or less, and then
> loses control due to wet or slippery roads and runs off the road

What the implementation does in contrast, is to manipulate the driving commands (like a tire puncture would do when driving). I am not sure, if this is what we should test here. In addition, there might be timing issues with this implementation: Assume you have two clients (the scenario + an AD stack). Both are sending steering commands to the CARLA server. The behavior may be different, depending on which command arrives first...

germanros1987:
I think reviewable got crazy. I did not approve.

fabianoboril:
And, the criteria should be updated to capture, that the vehicle re-gained control (i.e. came back to correct lane, drives normally, etc.). I suggest to do check the most simple things for now:
- No crash
- At the end of the scenario in correct lane (can be checked via waypoint) --> Add a new criterion for this
   (You could even count the number of ticks, the ego vehicle is not within the correct lane, and add a threshold to define success/failure)

fabianoboril:
@SubhashiniPerumal Thx for the updates. I tested it, and it looks much better than before. Still, I am bit worried about the usability of this scenario. It happened a few times, that the vehicle crashed against street lights, while the control loss was active. If we think about German's idea for the challenge, the scenario will always "win" against the client/user driving commands. So, the car crashes, because of the scenario, not because the user code could not handle the situation. In other words, the scenario will cause the failure, not the user code.

@germanros1987, @SubhashiniPerumal : Maybe we need to move this scenario to Town03, where the streets are wider? Or the control_loss needs to be updated, such the car doesn't hit a street light...
Suggestions?

fabianoboril:
Tested and approved. Can be merged.

PR #9, opened 2018-12-11
Feature/waypoint controller

germanros1987:
- Lateral control based on PID
- Longitudinal control based on PID
- VehicleControl class based on the combination of lateral and longitudinal control
- LocalPlanner class that set the right parameters given a target cruise speed
- AgentRoaming navigation: basic agent that navigates around a scene making random choices and respecting other vehicles and traffic lights

### DIFF
--- a/Navigation/controller.py
+++ b/Navigation/controller.py
@@ -1,0 +1,199 @@
+import datetime
+from collections import deque
+
+import math
+import numpy as np
+
+import carla
+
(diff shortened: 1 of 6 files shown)


fabianoboril:
I would like to test it, before giving my final approval...

fabianoboril:
Only one little thing remaining. Besides, the driving is a bit like a drunken guy... Maybe something we can work on, in future :)

PR #12, opened 2018-12-17
Implementation of traffic scenario 4

mohit-awana:
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/12)

### DIFF
--- a/Scenarios/object_crash_intersection.py
+++ b/Scenarios/object_crash_intersection.py
@@ -28,7 +28,7 @@ class VehicleTurningRight(BasicScenario):
     Location: Town01
     """
 
-    timeout = 80
+    timeout = 90
 
     # ego vehicle parameters
     _ego_vehicle_model = 'vehicle.nissan.micra'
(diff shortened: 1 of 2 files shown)


fabianoboril:
@germanros1987 : For me turning left works well. @mohit-awana Maybe adjust the trigger for the cyclist to not start too early (also take into account that the ego vehicle may wait at the red traffic light!)

@mohit-awana: I would recommend to adjust the position of the cyclist, for turning right. Move it closer to the intersection.

@germanros1987 , @mohit-awana : Regarding the success criteria: I think they are chosen well. Yet, I understand the point that it is already a success if the ego vehicle doesn't hit the cyclist, and if it continues driving it is a "bonus". I see to possibilities to add this to the analysis:
1. Currently the result of a criterion is pass/fail. However, we could add a third value (like a traffic light): good/ok/bad. Then we can differentiate, if a behavior was good (ego stopped in front of the cyclist, and then continued driving), ok (ego only stopped, but did not continue), or bad (crash).
2. We could introduce optional criteria: If a criterion is flagged as optional, it is not taken into consideration for the overall success analysis (success/failure of the entire scenario), but is still printed in the analysis.
We can also add both? What do you think? I can quickly add these, if you think it makes sense.

fabianoboril:
Please update Docs/list_of_supported_scenarios.md!

mohit-awana:
Accidentally deleted the local repository so hereby raising a new pull request for implementation of scenario 4 with all the required changes.

PR #25, opened 2019-01-23
Added BasicAgent as atomic behavior and updated documentation

fabianoboril:
Checklist:
  - [x] Your branch is up-to-date with the `master` branch and tested with latest changes
  - [x] Extended the README / documentation, if necessary
  - [x] Code compiles correctly and runs
  - [x] Code is formatted and checked with Utilities/code_check_and_formatting.sh

  * **Platform(s):**  Ubuntu 16.04
  * **Python version(s):** Python 2.7 and 3.5
  * **Unreal Engine version(s):** 4.19
  * **CARLA version:** 0.9.2

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/25)

### DIFF
--- a/ScenarioManager/atomic_scenario_behavior.py
+++ b/ScenarioManager/atomic_scenario_behavior.py
@@ -16,6 +16,8 @@ The atomic behaviors are implemented with py_trees.
 
 import py_trees
 import carla
+from agents.navigation.roaming_agent import *
+from agents.navigation.basic_agent import *
 
 from ScenarioManager.carla_data_provider import CarlaDataProvider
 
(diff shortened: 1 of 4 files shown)


ScenarioManager/atomic_scenario_behavior.py
```diff
@@ -666,3 +668,48 @@ def update(self):
         self._vehicle.apply_control(self._control)
 
         return new_status
+
+
+class BasicAgentBehavior(AtomicBehavior):
```
pravinblaze: The behaviour works fine for me. I tested scenario 9 to use this behaviour to execute turn. It behaves as expected.

Since we are trying to reduce location coupled parameters, is it ok to be using target location ? May be it can take a list of directives (left, right, straight, etc.) to execute like how the basic agent consumes [global route plan](https://github.com/carla-simulator/carla/pull/1174) ?

fabianoboril: As soon as your PR is merged, we can update the behavior to use turn information. I would see this as another PR.

PR #31, opened 2019-01-30
Added multiple loactions for Scenario 01

SubhashiniPerumal:
Checklist:

  - [X] Your branch is up-to-date with the `master` branch and tested with latest changes
  - [ ] Extended the README / documentation, if necessary
  - [X] Code compiles correctly and runs
  - [X] Code is formatted and checked with Utilities/code_check_and_formatting.sh
  - [ ] Changelog is updated

Description

- Added multiple locations for Traffic Scenario 01.

- Corrected the 'z' coordinate of target location as altitude loss is zero in Town 01.

Where has this been tested?

Platform(s): 18.04
Python version(s): 2.7
Unreal Engine version(s): ...
CARLA version: ... 0.9.3

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/31)

### DIFF
--- a/Docs/list_of_scenarios.md
+++ b/Docs/list_of_scenarios.md
@@ -3,13 +3,6 @@
 Welcome to the ScenarioRunner for CARLA! This document provides a list of all
 currently supported scenarios, and a short description for each one.
 
-## Scenarios in Town01
-The following scenarios take place in Town01. Hence, the CARLA server has to be
-started with Town01, e.g. when not using manual-mode:
-```
-CarlaUE4.sh /Game/Carla/Maps/Town01 -benchmark -fps=20 -windowed
-```
-
 ### FollowLeadingVehicle
 The scenario realizes a common driving behavior, in which the user-controlled
 ego vehicle follows a leading car driving down a given road in Town01. At some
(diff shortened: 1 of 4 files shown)


fabianoboril:
The second scenario in town01 does not work for me. It somehow does not terminate (it always runs into a timeout). Could you please check, if it properly executes for you?

nishanthjois:
@fabianoboril Second scenario in Town 01 is working for me i.e., the scenario ends when ego_vehicle reaches <y="160">.

fabianoboril:
Ah, I now why it was not working. I fixed the speed issue I reported in our meeting on Tuesday. As a result, the car was much faster, and passed then end point, while the jitters were still active.  I had to drive reverse, to get to the finish point.

--> Maybe increase the distance for the final checkpoint.

SubhashiniPerumal:
@fabianoboril:
- Updated the locations for Town04 and Town05 in config file.
- Query (this might need changes in control_loss.py and not in config file):
For Example: In Town03, Scenario 2 (ControlLoss_6):
After jitter action if we change the lane and do not return to the correct lane, scenario is not ending until timeout happens [If come back to correct lane then there is no issue].

Do you have any feedback on this?

SubhashiniPerumal:
Fixed the above lane issue by increasing the distance between vehicle and target location.

fabianoboril:
Please integrate the results of our last discussion, i.e. limit the jitter to the "left" side of the car (i.e. car moves only to the left, but not to the right)

SubhashiniPerumal:
I think we need not have additional logic for maintaining the speed. It's already in control. 
Kindly check and update if any changes needed.

fabianoboril:
Please also update the documentation (changelog)

zaaramariyam:
Closing this PR for now.

PR #34, opened 2019-01-31
Fixed velocity issue for Steering atomic

fabianoboril:
The current throttle/brake values are now maintained.

Checklist:
  - [x] Your branch is up-to-date with the `master` branch and tested with latest changes
  - [x] Extended the README / documentation, if necessary
  - [x] Code compiles correctly and runs
  - [x] Code is formatted and checked with Utilities/code_check_and_formatting.sh
  - [x] Changelog is updated

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/34)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## Lastest changes
+## Latest changes
 
 * Reworked scenario execution
     - Every scenario has to have a configuration provided as XML file.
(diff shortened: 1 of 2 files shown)


fabianoboril:
Ah, I now why it was not working. I fixed the speed issue I reported in our meeting on Tuesday. As a result, the car was much faster, and passed then end point, while the jitters were still active.  I had to drive reverse, to get to the finish point.

--> Maybe increase the distance for the final checkpoint.

PR #42, opened 2019-02-18
Jenkinsfile testing for scenario runner

akki6843:
#### Description
This is test version of Jenkins based pipeline for scenario runner, which will help in continuous integration.
As per the initial attempt it is expected from this file that it should return the python version.
Result of this test will help in determining the further step of either rework on pipeline code or work on sh file that will contain the commands to be executed automatically.  

Fixes #  
Results of this attempt will lead to fixes

Needs to be tested on Jenkins server

#### Possible Drawbacks
Since it will be tested in Jenkins server, if anything goes wrong check the Jenkins server 1st.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/42)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,11 @@
+pipeline{
+    agent { docker {image 'python:2.7.15'}}
+    stages{
+        stage('scenario runner test'){
+            steps{
+                sh 'python --version'
+            }
+
+        }
+    }
+}


fabianoboril:
Hm... Maybe we should discuss the testing procedure, before going into writing Jenkins scripts? How will a scenario be executed? What is success and what is failure? What docker image do we need to use, how do we handle different branches, etc.?

akki6843:
> Hm... Maybe we should discuss the testing procedure, before going into writing Jenkins scripts? How will a scenario be executed? What is success and what is failure? What docker image do we need to use, how do we handle different branches, etc.?

Sure, Lets discuss !

PR #45, opened 2019-02-19
Features/carla challenge evaluation

germanros1987:
#### Description

Provisional engine to run the CARLA challenge based on traffic scenarios 
Fixes #  

  * **Python version(s):** 3.5.5
  * **Unreal Engine version(s):** NA
  * **CARLA version:** 0.93

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/45)


felipecode:
___
*[srunner/scenarios/config_parser.py, line 91 at r1](https://reviewable.io/reviews/carla-simulator/scenario_runner/45#-LZCF1S188yHIkZz7R3U:-LZCF1S188yHIkZz7R3V:b-w1y5e4) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/59a98be83fbf3e9bf06d8b5e68c98174241f9e5a/srunner/scenarios/config_parser.py#L91)):*
> ```Python
> 
> 
> def parse_scenario_configuration(file_name, scenario_name):
> ```

This is breaking the scenario runner main loop.

felipecode:
Some comments:

1.  When running the challenge evaluator scenarios it is not clear that  the "group:" is used
even for a single XML file that may contain several scenarios . I would either make it clear how to use without the "group:" when executing a single scenario( Is it posible to do so ?) or maybe just remove the keyword all together. This way you would already execute everything from a XML.

2. the --list option is not working because of the required parameters.

3. The challenge  basic scenarios seem to be working ! all of them.

4. This repository is called "scenario runner" but i think that now it is becoming mainly the CARLA challenge repo, is this going to reflect on the name of the repository ? That would help people to know clearly where the challenge material is.

felipecode:
___
*[scenario_runner.py, line 20 at r1](https://reviewable.io/reviews/carla-simulator/scenario_runner/45#-LZCMlYp7VJKvXr9x0hP:-LZCMlYp7VJKvXr9x0hQ:babexv) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/59a98be83fbf3e9bf06d8b5e68c98174241f9e5a/scenario_runner.py#L20)):*
> ```Python
> from argparse import RawTextHelpFormatter
> from datetime import datetime
> ```

I guess Scenario runner would be missing the same addition of the CARLA_ROOT path now.

fabianoboril:
> Some comments:
> 
> 1. When running the challenge evaluator scenarios it is not clear that  the "group:" is used
>    even for a single XML file that may contain several scenarios . I would either make it clear how to use without the "group:" when executing a single scenario( Is it posible to do so ?) or maybe just remove the keyword all together. This way you would already execute everything from a XML.
> 4. This repository is called "scenario runner" but i think that now it is becoming mainly the CARLA challenge repo, is this going to reflect on the name of the repository ? That would help people to know clearly where the challenge material is.

@1: The intent is that you can run a single scenario only, which is required for everyone seriously testing CARLA with an AD stack. Therefore, I strongly recommend to keep the differentiation to execute a single scenario or a complete scenario list from a file. This was also possible before the last rounds of commits. I will check it today, if things are still working.
@4: This repo is intended to keep the everything required for scenario execution (at least this was my understanding). If there are particular things for the challenge (I haven't see these yet), then we should maybe move them out. I would vote against moving the scenario runner, as people started using it for various reasons (and the challenge is only one of them).

fabianoboril:
After rebasing we can merge it from my side.

PR #51, opened 2019-02-25
Pseudo velocity sensor

felipecode:
#### Description

It is reasonable for the agents to also have access to its current forward speed. 
However, speed, right now is not defined as a sensor, but as part of the vehicle
transformation.
This PR adds a "sensor", coded in python, that asynchronously provide speeds.
The sensor can be add to an agent as follows:

                   ['sensor.speedometer', {'reading_frequency': 25},
                    'speed'
                    ],

Also added a traceback to print the full exception for the scenario manager.
line 234. I think the exception catch there is too broad and make debug a bit
hard.

  * **Platform(s):**ubuntu 16.04
  * **Python version(s):**  3.5
  * **CARLA version:**  CARLA 0.9.3

#### Possible Drawbacks

One problem is that we have to set a sampling rate that normally is different than the sampling rate of the other sensors collected in CARLA.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/51)

### DIFF
--- a/scenario_runner.py
+++ b/scenario_runner.py
@@ -17,6 +17,7 @@ from __future__ import print_function
 import argparse
 from argparse import RawTextHelpFormatter
 from datetime import datetime
+import traceback
 
 import sys
 
@@ -234,6 +235,7 @@ class ScenarioRunner(object):
                                               args.debug)
                 except Exception as exception:
                     print("The scenario cannot be loaded")
+                    traceback.print_exc()
                     print(exception)
                     self.cleanup()
                     continue
--- a/srunner/challenge/autoagents/DummyAgent.py
+++ b/srunner/challenge/autoagents/DummyAgent.py
@@ -28,7 +28,6 @@ class DummyAgent(AutonomousAgent):
         """
         sensors = [{'type': 'sensor.camera.rgb', 'x':0.7, 'y':0.0, 'z':1.60, 'roll':0.0, 'pitch':0.0, 'yaw':0.0,
                     'width':800, 'height': 600, 'fov':100, 'id': 'Center'},
-
                    {'type': 'sensor.camera.rgb', 'x':0.7, 'y':-0.4, 'z': 1.60, 'roll': 0.0, 'pitch': 0.0,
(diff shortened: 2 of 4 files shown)


felipecode:
___
*[srunner/challenge/challenge_evaluator.py, line 176 at r1](https://reviewable.io/reviews/carla-simulator/scenario_runner/51#-LZf07HO76dl0FFkFDvL:-LZf1Rn2Dhli_utiXQWr:b-896fix) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/57e91a4927aa6832db75ab9ced3e0e40d1a1bcdc/srunner/challenge/challenge_evaluator.py#L176)):*
<details><summary><i>Previously, germanros1987 wrote…</i></summary><blockquote>

looks good
</blockquote></details>

Done.

felipecode:
___
*[scenario_runner.py, line 238 at r1](https://reviewable.io/reviews/carla-simulator/scenario_runner/51#-LZf028E8cqf6wVBIunM:-LZf0xuFEfMleGe-IF0j:b-1e7vyx) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/57e91a4927aa6832db75ab9ced3e0e40d1a1bcdc/scenario_runner.py#L238)):*
<details><summary><i>Previously, germanros1987 wrote…</i></summary><blockquote>

Is this really supposed to be here...?
</blockquote></details>

I can remove. We should maybe not catch so broad exceptions. At least with the traceback I know what happened. Otherwise it just printed the last error.

felipecode:
___
*[scenario_runner.py, line 238 at r1](https://reviewable.io/reviews/carla-simulator/scenario_runner/51#-LZf028E8cqf6wVBIunM:-LZf2kBV--qcbNnQssSU:b-896fix) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/57e91a4927aa6832db75ab9ced3e0e40d1a1bcdc/scenario_runner.py#L238)):*
<details><summary><i>Previously, germanros1987 wrote…</i></summary><blockquote>

If you think it helps we can keep it. Your call. 
</blockquote></details>

Done.

PR #101, opened 2019-03-21
Updated TS 01

mohit-awana:
#### Description
*Added sequence of perturbations with some time gap.

  * **Platform(s):**  Ubuntu 18.04
  * **Python version(s):** 2.7
  * **Unreal Engine version(s):** 4.21
  * **CARLA version:** 0.9.4

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/101)


mohit-awana:
@germanros1987 Right now the scenario concludes on encountering an intersection, should we have a different end condition?

PR #107, opened 2019-03-22
Rework of TS07

fabianoboril:
Completed rework of scenario 07 according to feedback. We did not use a stream of adversaries for now, to increase the challenge of this scenario, but we modified the behavior to ensure a collision for various driving behaviors.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/107)

### DIFF
--- a/srunner/scenarios/opposite_vehicle_taking_priority.py
+++ b/srunner/scenarios/opposite_vehicle_taking_priority.py
@@ -15,6 +15,7 @@ import sys
 
 import py_trees
 import carla
+from agents.navigation.local_planner import RoadOption
 
 from srunner.scenarios.basic_scenario import *
 from srunner.scenarios.scenario_helper import *
@@ -131,9 +132,8 @@ class OppositeVehicleRunningRedLight(BasicScenario):
 
         sync_arrival = SyncArrival(
             self.other_actors[0], self.ego_vehicle, location_of_collision_dynamic)
-        sync_arrival_stop = InTriggerDistanceToVehicle(self.other_actors[0],
-                                                       self.ego_vehicle,
-                                                       41)
+        sync_arrival_stop = InTriggerDistanceToNextIntersection(self.other_actors[0],
+                                                                5)
         sync_arrival_parallel.add_child(sync_arrival)
         sync_arrival_parallel.add_child(sync_arrival_stop)
 
@@ -159,7 +159,7 @@ class OppositeVehicleRunningRedLight(BasicScenario):
             policy=py_trees.common.ParallelPolicy.SUCCESS_ON_ONE)
 
         continue_driving_waypoints = WaypointFollower(
-            self.other_actors[0], self._other_actor_target_velocity, plan=plan)
+            self.other_actors[0], self._other_actor_target_velocity, plan=plan, avoid_collision=False)
 
         continue_driving_distance = DriveDistance(
             self.other_actors[0],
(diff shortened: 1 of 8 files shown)


srunner/scenarios/opposite_vehicle_taking_priority.py
```diff
@@ -133,24 +133,44 @@ def _create_behavior(self):
...
                                                        self.ego_vehicle,
-                                                       15)
+                                                       41)
         sync_arrival_parallel.add_child(sync_arrival)
         sync_arrival_parallel.add_child(sync_arrival_stop)
 
-        keep_velocity_for_distance = py_trees.composites.Parallel(
-            "Keep velocity for distance",
+        # Generate plan for WaypointFollower
+        waypoint = self.other_actors[0].get_world().get_map().get_waypoint(self.other_actors[0].get_location())
```
mohit-awana: get_map() is now available in CarlaDataProvider.

srunner/scenarios/opposite_vehicle_taking_priority.py
```diff
@@ -133,24 +133,44 @@ def _create_behavior(self):
...
             name="Distance")
-        keep_velocity_timeout = TimeOut(5)
-        keep_velocity_for_distance.add_child(keep_velocity)
-        keep_velocity_for_distance.add_child(keep_velocity_distance)
-        keep_velocity_for_distance.add_child(keep_velocity_timeout)
+
+        continue_driving_timeout = TimeOut(10)
+
+        continue_driving.add_child(continue_driving_waypoints)
+        continue_driving.add_child(continue_driving_distance)
```
mohit-awana: Currently, if the ego vehicle decides to drive in the same direction as adversary, the ego vehicle is stuck behind the adversary vehicle, it has to change lane to complete the scenario. Is it intended? Otherwise scenario ends with a timeout as the end condition of the scenario doesn't meet.

srunner/scenarios/opposite_vehicle_taking_priority.py
fabianoboril: Fixed.

srunner/scenarios/opposite_vehicle_taking_priority.py
fabianoboril: Yes correct. The only chance I see to avoid this, would be to feed the ego route into the scenario. This however, would require some re-design.

For now, I think we can still use this scenario. Even if the adversary takes the same route, it will be in front of the ego vehicle (so a follow maneuver), and disappear as soon as this scenario is over. The disappearance could be made even a bit more aggressively.

So, all in all, my suggestion would be to keep it as it is for now. If our testing reveals that it is a real problem, we can come back and work on it.

fabianoboril:
@mohit-awana: Can we merge it?

mohit-awana:
> @mohit-awana: Can we merge it?

Okay, yes please :D

mohit-awana:
Sorry, my bad :D Pressed the wrong button :P

fabianoboril:
> > ```
> > OppositeVehicleRunningRedLight011
> > ```
> 
> I am not getting the expected behavior. Have a look:
> 
> https://youtu.be/vCH2CsEMj-Y
> 
> > _[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/107#-:-Lb9g3RH1Ib0VhmywJiW:b-mcxzk1)_ status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @germanros1987 and @mohit-awana)

Thanks for the video. The new default-behavior of avoiding collisions of the WaypointFollower caused the other actor to not move. I fixed this and also updated the behavior slightly to improve the scenario for low ego velocities.

fabianoboril:
Rebase done :)

PR #108, opened 2019-03-22
Feat trigger location scenario

### LINKED ISSUE #106: Wrong behavior in object_crash_vehicle scenario
We have detected an error in object_crash_vehicle

The reference to derive the location of the adversaries is obtained via self.ego_vehicle. 

```
lane_width = self._reference_waypoint.lane_width
location, _ = get_location_in_distance(self.ego_vehicle, _start_distance)
waypoint = self._wmap.get_waypoint(location)
```

However, this is incorrect for all those cases where the ego_vehicle is still far away from the scenario. The correct object would be config.ego_vehicle, which despite its horribly misleading name refers to the trigger position of the scenario.

I would advise to:

 * Change the parser to assign the right name to this field, i.e., scenario_trigger_location
* Fix this behavior

germanros1987:
#### Description

This PR updates the reference location of scenarios, that now is called trigger_point. It also fixes #106 

Fixes #106

  * **Python version(s):** 3.5.5
  * **Unreal Engine version(s):** NA
  * **CARLA version:** overnight 2019-03-19

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/108)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## Latest changes
+* Updated the reference position for a scenario is now called trigger_point
 * Added universal access to the map without re-calling get_map()
 * Added criteria_enable flag to enable/disable criteria tree
 * Added multiple helper methods for generic scenario execution.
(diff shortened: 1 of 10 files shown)


felipecode:
:lgtm:

felipecode:
:lgtm:

PR #150, opened 2019-04-05
Add transform based trigger point check

felipecode:
#### Description
Instead of checking time of arrival and trigger activation by location we
do it by also setting a threshold on the angle.

  * **Platform(s):** ...
  * **Python version(s):** 3.5
  * **CARLA version:** 9.5

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/150)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Latest changes
 * Added background scenario to generate dynamic traffic using autopilot
+* Comparisons to the trigger are made with transform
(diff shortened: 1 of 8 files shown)


fabianoboril:
With this PR you may break many scenarios, which are now only relying on locations and not using orientation. Is this required for the challenge, or is there another way (maybe just add new behaviors, or make using the orientation optional). Otherwise, this may throw us weeks back and we need to start testing single scenario sequences...

PR #163, opened 2019-04-08
TS3 and TS4 Cleaning

felipecode:
#### Description

Several spawn positions were not happening for scenarios 3 and  4. The problem
is that the spawn position was blocked, and that depends on the trigger and on
the distance to the trigger the scenarios are spawned.

Also
Removed the physics, so the vehicles dont fall into the infinite on the beggining.
Fixed a wrong behavior from scenario 4 selection

Removed bad start positions

  * **Platform(s):** ...
  * **Python version(s):** 3.5 
  * **CARLA version:** 9.5

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/163)


srunner/scenariomanager/atomic_scenario_behavior.py
```diff
@@ -1120,6 +1121,7 @@ def update(self):
         self._actor.set_velocity(carla.Vector3D(0, 0, 0))
         self._actor.set_angular_velocity(carla.Vector3D(0, 0, 0))
         self._actor.set_transform(self._transform)
+        self._actor.set_simulate_physics(enabled=True)
```
nishanthjois: I tested Control Loss with this change: Due to this change, debris is above the ground and causes ego-vehicle to crash. 
Screenshot from 2019-04-08 15-47-51

felipecode: Good ! I think I noticed it once ! Good catch.

felipecode: Fixed

PR #182, opened 2019-04-10
Set range of Lidar to 50m

fpasch:
Lidar range is in centimeters. Therefore 2m is not very helpful.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/182)


fpasch:
I also updated the documentation.
https://github.com/carla-simulator/carla/pull/1523

PR #214, opened 2019-04-13
 calculate the distance along route TS4

felipecode:
#### Descriptio

Use distance along route to activate scenario 04

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/214)


felipecode:
Done

PR #220, opened 2019-04-15
Implement benchmarks in style of CoRL 2017 / Carla 100

AdamGleave:
#### Description

The benchmarks currently implemented in the Scenario Runner are focused on the Carla Challenge and are both considerably more realistic and challenging than previous benchmarks such as CoRL 2017. While this is overall a good thing, it is valuable to be able to directly compare agents developed for the Carla Challenge or other purposes with previous benchmarks. This validates that the new benchmarks are in fact harder than the previous, and lets researchers perform regression tests when porting algorithms previously validated on the old benchmarks in the 0.8.x branch.

This PR aims to port the CoRL 2017 and Carla 100 tests in https://github.com/carla-simulator/driving-benchmarks for the 0.8.x branch to Scenario Runner on the 0.9.x branch. An exact reproduction will likely not be possible due to some differences in setup, but I intend to keep it as close as possible and clearly document any differences.

This is a WIP. Tasks:
  - [X] Generate route XML files.
  - [x] Implement appropriate termination conditions (reaching goal or dynamic timeout)
  - [x] Make volume and style of background vehicle/pedestrian traffic match
  - [x] Handle weather.
  - [x] Report metrics. (Optional: may not want to match the original benchmarks exactly, as the scenario runner supports a richer set of metrics.)
 - [x] Validate conditional imitation learner baseline on new benchmarks and verify it is similar to results on old benchmark for 0.8.x.

  * **Python version(s):** 3.5
  * **Unreal Engine version(s):** UE4
  * **CARLA version:** 0.9.5

#### Possible Drawbacks

Directly porting the code https://github.com/carla-simulator/driving-benchmarks to 0.9.x would allow for a more faithful reproduction of the original benchmark. However, some differences seem inevitable in any case due to 0.8.x vs 0.9.x changes, and so overall it seems better to me to focus development efforts for 0.9.x in scenario_runner.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/220)


AdamGleave:
I'm intending to focus on this for the next few days so would appreciate feedback on the roadmap. Particular areas:
  - Is this something other people are interested in? (I need something like this anyway, but if I'm the only user, I'll make different design decisions.)
  - To handle the different termination conditions, I'm planning on defining a new class `BenchmarkMasterScenario` class that fills a similar role to `MasterScenario`, and making which one is used configurable depending on the experiment. Does this seem sensible?
  - For background vehicle/pedestrian traffic, I'm planning on making this configurable at a much finer grained level so it can be matched, but otherwise keep `BackgroundActivity` basically the same.
  - Any ideas on other major differences between `scenario_runner` and `driving-benchmarks` that I've missed?

AdamGleave:
I'm most of the way through this. The main thing I've not been able to match is the pedestrian traffic since Carla 0.9.5 doesn't have an autopilot for pedestrians; it should be easy to add this later though. I'll focus on code cleanup and testing tomorrow; any feedback on the design would be appreciated.

felipecode:
Hello Adam ! I am really interested on this feature. I am unfortunately a bit busy with the challenge and other projects to be able to help much. 
The missing pedestrians would make the benchmarks to not match, for sure. We are expecting to have this feature on the next releases. Some other notes:

* The background scenario has something called traffic jam control that remove cars that are
creating jams, this was not present on the previous benchmarks.
* I think the result reporting of the scenario runner is much more sensible, however it misses doing experiments separately. That is a very useful feature. On the benchmark every single experiment is logged and you can do them in parallel or continue from any experiment if any crashes happen.
* The definition of a experiment is richer at the benchmarks, since it encapsulates metrics not only routes and scenarios, that would be something interesting to exist.  Feel free to ping me at discord if you want to discuss anything.

What you already made seems great ! I am definitely using it in the future. Probably it would be cool to update the benchmarks repo to have Corl2017, Carla100 CarlaADchallenge2019 and others and keep all of the them with some common interface and depending on the scenario runner.  I am not sure how to integrate it in this repo. We will probably discuss this after the challenge.

AdamGleave:
Thanks for the feedback. I had indeed overlooked the traffic jam remover. I'll make that a configurable option and disable it by default in CoRL 2017 / Carla 100.

I'll look into how difficult it would be for me to add a feature to run experiments in parallel / checkpointing. This sounds like an area that might require modifying a lot of the scenario runner internals, in which case it might be more your comparative advantage. In any case it seems nice to have but not urgent as it's primarily a performance issue.

I'll have a think about ways to make metrics configurable, seems like it shouldn't be too hard to specify what set of metrics we care about, might be harder to control them at a more fine-grained level.

Definitely in favour of making them all a common interface. Happy to discuss this with you once you have more bandwidth.

AdamGleave:
This PR is now ready for review (although I understand if you don't have time right now).

In addition to porting the benchmark, I've improved exception handling / logging (I found previously the scenario runner would often silently fail) and added options to control the spectator camera for interactive debugging purposes.

The main features that are missing still are:
  - Pedestrians. Easy to add once it's supported in Carla 0.9.x (just delete two lines of code that skip 'walker.*').
  - Matching the metrics supported in the old benchmark suite. I think the scenario runner overall has better metrics so I'm not planning on putting the time in to do this, but it would be nice to have (or more configurable metrics, as you suggest).

Additionally, I tested the conditional imitation learning baseline with this and found significantly worse results than reported in the CoRL 2017 paper. I believe this is accurate since it matches the qualitative performance I've observed. It'd be nice to get imitation learning & RL baselines that are similarly as strong as in the CoRL 2017 paper, so that's likely to be the next thing I work on.

felipecode:
Do you know why you have issues reproducing the results ?
The results should be better not worse, since we don't have the pedestrians.
I can try to help you reproducing to try to figure out the mistake.

I am now working in the repository to integrate all the benchmarks and the data collector in a single system.
It is not open yet, it will be called: CARLA experience generator (C-EX). I am using part what you generated in this PR. I can invite you to it soon !

The basic idea is that you have an exp object that can be generated by an exp generator.
```
c_ex = C_EX(json_config, docker, number_of_iterations, params['batch_size'])
c_ex.start()  # Will start all the docker carla servers specified on batch size.
for env in c_ex:  # C_EX has an iterator that produces environments
	states, rewards = agent.unroll(env) # unroll a batch of episodes
	agent.reinforce(rewards) # reinforce the results of this 
        env.get_summary() # Get the summary of the experience ( beenchmark)

```

The environment, and its instance, an experience, is a set of scenarios and route as well as an evaluation criteria ( to be added) that you can run by running successive steps.

```
while exp.is_running():
       state, reward = exp.run_step(controls)  # This already saves data and check benchmark in parallel
```

AdamGleave:
> Do you know why you have issues reproducing the results ?
> The results should be better not worse, since we don't have the pedestrians.
> I can try to help you reproducing to try to figure out the mistake.
I think the issue is graphics & dynamics changes between 0.8 and 0.9. When I used https://github.com/felipecode/coiltraine/blob/master/view_model.py it seemed notably less reliable, e.g. if I added other vehicles to the simulation it'd often crash into them. I didn't actually try the same model on 0.8 though, so it's possible there are other differences between https://github.com/felipecode/coiltraine/blob/master/view_model.py and https://github.com/carla-simulator/imitation-learning that are important.

I'm sure there are also some differences in the benchmark I developed that will need to be fixed, but I expect they'll cause single-digit % change not double-digit %.

> I am now working in the repository to integrate all the benchmarks and the data collector in a single system.
> It is not open yet, it will be called: CARLA experience generator (C-EX). I am using part what you generated in this PR. I can invite you to it soon !
Great, happy to collaborate on this, although will be a bit time constrained in the next few weeks (NeurIPS deadline).

> The basic idea is that you have an exp object that can be generated by an exp generator.
> 
> ```
> c_ex = C_EX(json_config, docker, number_of_iterations, params['batch_size'])
> c_ex.start()  # Will start all the docker carla servers specified on batch size.
> for env in c_ex:  # C_EX has an iterator that produces environments
> 	states, rewards = agent.unroll(env) # unroll a batch of episodes
> 	agent.reinforce(rewards) # reinforce the results of this 
>         env.get_summary() # Get the summary of the experience ( beenchmark)
> ```
I'm a little unclear on the interface here. Is `agent.unroll` inference and `agent.reinforce` training? This seems fine for policy gradient methods that have larger batch sizes but would preclude use of most DQN methods that update after a small number of timesteps.

For RL purposes I think ideally we'd be able to provide a `Gym` interface, and make it easy to make a `VecEnv` (from `baselines`) corresponding to different Carla servers. The fact Carla is more of an environment generator (different maps, non-player characters, weather conditions, etc) makes it harder to fit into that framework but I think is doable.

felipecode:
> I think the issue is graphics & dynamics changes between 0.8 and 0.9. When I used https://github.com/felipecode/coiltraine/blob/master/view_model.py it seemed notably less reliable, e.g. if I added other vehicles to the simulation it'd often crash into them. I didn't actually try the same model on 0.8 though, so it's possible there are other differences between https://github.com/felipecode/coiltraine/blob/master/view_model.py and https://github.com/carla-simulator/imitation-learning that are important.
> I'm sure there are also some differences in the benchmark I developed that will need to be fixed, but I expect they'll cause single-digit % change not double-digit %.
The problem is that the  https://github.com/carla-simulator/imitation-learning is a drastically different agent than the challenge baseline. They use very different data and CARLA versions of data. It would like to try to replicate the published versions.



> I'm a little unclear on the interface here. Is `agent.unroll` inference and `agent.reinforce` training? This seems fine for policy gradient methods that have larger batch sizes but would preclude use of most DQN methods that update after a small number of timesteps.
> 
> For RL purposes I think ideally we'd be able to provide a `Gym` interface, and make it easy to make a `VecEnv` (from `baselines`) corresponding to different Carla servers. The fact Carla is more of an environment generator (different maps, non-player characters, weather conditions, etc) makes it harder to fit into that framework but I think is doable.

You are completely right. The repo is more about the environment generation. The agent interface
on this example is more for a policy gradient based method but definitely should be expanded for other approaches.
It is interesting to merge with gym ! However I am currently not intending to do so.

PR #222, opened 2019-04-15
Fix: Setting control for TS01 when running outside challenge route 

akki6843:
Check if control loss scenario is running independently or is a part of challenge route

* Modified ChangeNoiseParameters behaviour, so that control is applied via behaviour  if not a part of challenge route.

#### Fix #  
Throttle and steering were not getting applied when we run TS01 outside challenge route

  * **Python version(s):** 3.5
  * **CARLA version:** 9.5

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/222)


fabianoboril:
I close this PR for now. If we decide to come back to this update, we can just re-open it.

PR #226, opened 2019-04-16
Adding InTriggerDistanceToLocationAlongRoute behaviour to the tree in TS03

akki6843:
Adding InTriggerDistanceToLocationAlongRoute behaviour to the tree
* removed timeout from the tree
* added names to the leaf nodes for better log entry

  * **Python version(s):** 3.5
  * **CARLA version:** 0.9.5

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/226)

### DIFF
--- a/srunner/scenarios/object_crash_vehicle.py
+++ b/srunner/scenarios/object_crash_vehicle.py
@@ -157,6 +157,8 @@ class DynamicObjectCrossing(BasicScenario):
         # Number of attempts made so far
         self._spawn_attempted = 0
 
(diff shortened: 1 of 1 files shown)


felipecode:
TS 3 is fine, with a just a few places where maybe the pedestrians would spawn too early.
But i think it makes complete sense to depend on time to arrival. Has this been tested on several routes ? 40+ routes  on all towns ? what is the advantage ? Is there any bug associated with the previous approach ?

PR #241, opened 2019-05-02
Control loss fix low speeds

felipecode:
#### Description

The control loss function now only adds noise if the velocity is bigger than threshold.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/241)

### DIFF
--- a/srunner/scenarios/control_loss.py
+++ b/srunner/scenarios/control_loss.py
@@ -140,6 +140,7 @@ class ControlLoss(BasicScenario):
                                                      policy=py_trees.common.ParallelPolicy.SUCCESS_ON_ONE)
         # Abort jitter_sequence, if the vehicle is approaching an intersection
         jitter_abort = InTriggerDistanceToNextIntersection(self.ego_vehicle, self._abort_distance_to_intersection)
+
         # endcondition: Check if vehicle reached waypoint _end_distance from here:
         end_condition = DriveDistance(self.ego_vehicle, self._end_distance)
         start_end_parallel.add_child(start_condition)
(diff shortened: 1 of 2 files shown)


felipecode:
___
*[srunner/scenarios/control_loss.py, line 27 at r1](https://reviewable.io/reviews/carla-simulator/scenario_runner/241#-LdwCMsb7cidrq95ZH2T:-Le9wE2d2Y-uxVxCkNSw:bn2qhxc) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/74f91da5aa8106920ac19c50e158e3857b84ab8e/srunner/scenarios/control_loss.py#L27)):*

Is this 5 m/s or 5 km/h? 

In addition, people may see that, and will drive slower to avoid this scenario.... I am not sure if this is a good idea!
</blockquote></details>

You are right. But having noise when the car is very slow is bad. Maybe I should reduce to 2 m/s ?

felipecode:
___
*[srunner/scenarios/control_loss.py, line 27 at r1](https://reviewable.io/reviews/carla-simulator/scenario_runner/241#-LdwCMsb7cidrq95ZH2T:-Le9wMD88qS6ffJIa8_1:bv46ns1) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/74f91da5aa8106920ac19c50e158e3857b84ab8e/srunner/scenarios/control_loss.py#L27)):*
<details><summary><i>Previously, felipecode (Felipe Codevilla) wrote…</i></summary><blockquote>

You are right. But having noise when the car is very slow is bad. Maybe I should reduce to 2 m/s ?
</blockquote></details>

This is bad specially for throttle noise

fabianoboril:
2m/s is a better value, if you really need to limit it.

PR #316, opened 2019-07-06
Bugfix collision test criterion test status init

jit-rc:
#### Description

Fixed Criterion CollisionTest to update status to SUCCESS/FAILURE, it used to stay in INIT state for SUCCESS 

  * **Platform(s):** Ubuntu 16.04.6 LTS
  * **Python version(s):** 3.5.2
  * **CARLA version:** 0.9.5

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/316)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
       termination
     - Added behavior to activate/deactivate autopilot
     - Fixed WaypointFollower initialization
+* Fixes
+    - Fixed Criterion CollisionTest to update status to SUCCESS/FAILURE
 
 
 ## CARLA Scenario_Runner 0.9.5
--- a/srunner/scenariomanager/atomic_scenario_criteria.py
+++ b/srunner/scenariomanager/atomic_scenario_criteria.py
@@ -279,6 +279,11 @@ class CollisionTest(Criterion):
         """
         new_status = py_trees.common.Status.RUNNING
 
+        if self.actual_value > 0:
+            self.test_status = "FAILURE"
+        else:
+            self.test_status = "SUCCESS"
+
         if self._terminate_on_failure and (self.test_status == "FAILURE"):
             new_status = py_trees.common.Status.FAILURE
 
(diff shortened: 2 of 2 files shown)


fabianoboril:
Did you check the latest version on the master branch? We just worked a fix for this issue? Is it still there?

jit-rc:
> Did you check the latest version on the master branch? We just worked a fix for this issue? Is it still there?
Yes, it is not there in master, but this is a copy from older implementation, that was removed here https://github.com/carla-simulator/scenario_runner/commit/7164e51915df20867fc4936c5cae3ad3ec3e1e0d

jit-rc:
Without this change you can simple try FollowLeadingVehicle_1 with manual controls, it would always result in failure as the CollisionTest State remains in INIT when there is no crash, and !SUCCESS is treated as failure

fabianoboril:
> Without this change you can simple try FollowLeadingVehicle_1 with manual controls, it would always result in failure as the CollisionTest State remains in INIT when there is no crash, and !SUCCESS is treated as failure

I tested it, and the issue does not exist in the current master branch. Please note the changes we have done recently to the Criterion base class. The states RUNNING and INIT are mapped to SUCCESS.

```
class Criterion(py_trees.behaviour.Behaviour):

    """
    """

    def __init__(...):
        super(Criterion, self).__init__(name)
        self.logger.debug("%s.__init__()" % (self.__class__.__name__))
        self._terminate_on_failure = terminate_on_failure

        self.name = name
        self.actor = actor
        self.test_status = "INIT"

    ....

    def terminate(self, new_status):
        if (self.test_status == "RUNNING") or (self.test_status == "INIT"):
            self.test_status = "SUCCESS"
```

fabianoboril:
@jit-rc Please test again with plain master to confirm 👍

jit-rc:
Sorry I missed the change in the Base class, any plans to merge it is carla_challenge branch, or should we merge it locally or use master branch from the challenge, as it exists in the challenge, and that would be wrong evaluation of score. 

Also in the base class defaulting to  if ( "RUNNING")  = "SUCCESS", at terminate, does feel like a safe default to keep. As in many cases RUNNING at terminate would mean FAILURE. I think keeping the success or failure explicitly to the criterion implementation is a better idea, if there is need to keep a default it better to make FAILURE as default, at least someone would debug the FAILURE and not have SUCCESS by mistake, where someone forgot to handle termination properly

fabianoboril:
No worries.

Regarding branch usage: The first iteration of the challenge is now completed, so you can safely use the master branch. For the next iteration, we will most likely create a new branch or update the current one. This is not yet decided.

Regarding the default choice. This is of course a trade-off. There are pros and cons for both sides. At the end, everyone can overwrite the terminate() method and choose a different implementation. For everything that goes into this repo, our code reviews should ensure that things are consistent and work. Let's see if this works ;-)

PR #405, opened 2019-11-18
pip install/ setup.py

felipecode:
Adding a setup.py to be able to install this repo via pip.
Currently it is already installable through

     pip install srunner

  * **Platform(s):** ...
  * **Python version(s):** ...
  * **CARLA version:** ...

#### Possible Drawbacks

The upload to pip requires a new version every time you upload
we should be careful to only upload stable versions.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/405)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Latest changes
 
+* Added the setup.py file to upload to pip
 * Split of behaviors into behaviors and conditions
 * Moved atomics into new submodule scenarioatomics
 * Updated documentation for all behaviors, conditions and test criteria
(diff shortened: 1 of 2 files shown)


felipecode:
The new setup now installs all dependencies except CARLA. I also set manual control and scenario runner as entry scripts.
Doing a pip install for CARLA is a bit more tricky because it requires compiling cpp code. We have this already on the .egg but pip does not support .egg directly. . It seems there is already a draft there:
https://github.com/carla-simulator/carla/blob/master/PythonAPI/carla/setup.py
I will ask marc/bernard/nestor to assist me on that.

I uploaded another version to pypi

fabianoboril:
Outdated. Will close this PR for now.

PR #408, opened 2019-11-24
Fix agent import error

Yougmark:
#### Description

`self.agent_instance = getattr(self.module_agent, self.module_agent.__name)(args.config)` throws exception for not able to find module, if the agent is HumanAgent, RosAgent, DummyAgent, or NpcAgent. Because the filename and class name does not match.

Generate class name from `__name__`
    
Potential problem that this requires the Agent and file name matches in the following way:
    
    HumanAgent vs. human_agent.py
    NpcAgent vs. npc_agent.py
    
    CapitalizedAgent vs. capitalized_agent.py

Fixes #  

  * **Platform(s):**  Ubuntu 18.04.3 LTS
  * **Python version(s):** Python 2.7 and 3.5
  * **Unreal Engine version(s):** 4.22.3
  * **CARLA version:** latest (after 0.9.6)

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/408)

### DIFF
--- a/srunner/challenge/autoagents/dummy_agent.py
+++ b/srunner/challenge/autoagents/dummy_agent.py
@@ -83,4 +83,4 @@ class DummyAgent(AutonomousAgent):
 
         return control
 
-__name__="DummyAgent"
+__name__ = "DummyAgent"
--- a/srunner/challenge/autoagents/dummy_agent.py
+++ b/srunner/challenge/autoagents/dummy_agent.py
@@ -82,5 +82,3 @@ class DummyAgent(AutonomousAgent):
         control.hand_brake = False
 
         return control
-
-__name__ = "DummyAgent"
--- a/srunner/challenge/autoagents/dummy_agent.py
+++ b/srunner/challenge/autoagents/dummy_agent.py
@@ -82,3 +82,5 @@ class DummyAgent(AutonomousAgent):
         control.hand_brake = False
 
         return control
+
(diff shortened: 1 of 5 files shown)


Yougmark:
It's failing the pylint check with error "Redefining built-in '__name__' (redefined-builtin)".  Any suggestions?

fabianoboril:
Hi,
thank you for this PR. You can simply add a single line before the throw occurs with
`agent_class_name = self.module_agent.__name__.title().replace('_', '')`
This will convert for example human_agent to HumanAgent.

Yougmark:
> Hi,
> thank you for this PR. You can simply add a single line before the throw occurs with
> `agent_class_name = self.module_agent.__name__.title().replace('_', '')`
> This will convert for example human_agent to HumanAgent.

That'll work for now. I updated. But it requires the class name matches the file name in certain way.

fabianoboril:
> > Hi,
> > thank you for this PR. You can simply add a single line before the throw occurs with
> > `agent_class_name = self.module_agent.__name__.title().replace('_', '')`
> > This will convert for example human_agent to HumanAgent.
> 
> That'll work for now. I updated. But it requires the class name matches the file name in certain way.

The old challenge_evaluator has a replacement already (can be used from the scenario_runner directly) and for the new challenge we are about to create a new setup. Therefore, I think this workaround is fine.

PR #433, opened 2020-01-23
Glopez/scenario runner

glopezdiest:
#### Description

Added 3 new atomic criteria:
- CollisionTestLeaderboard which only takes into account collisions with the same object once
- WronglaneTestPerMeter: variant of WrongWayTest but it additionally calculates the distance driven through the wrong lane
- OnSidewalkTestPerMeter: variant of OnSidewalkTest but it additionally calculates the distance driven through the sidewalk

  * **Python version(s):** 3.6
  * **CARLA version:** 0.9.7

#### Possible Drawbacks

None, nothing was changed, just added

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/433)


glopezdiest:
yes, I basically left it open as I found new info after doing the pull_request (should have sent a message, sorry).

The idea is just to also add the amount of meters traveled through the sidewalk / wrong way, and the Collision one prevents the event from triggering multiple times from the same object. The first two could be modified but both their TrafficEvents have been changed to trigger after the car has returned to the lane, not when going outside of it. Will it affect any other areas? If not, I'll modify them instead of adding them.

fabianoboril:
You can modify the existing ones. There is a branch for the old challenge, if we ever need to go back ;)

glopezdiest:
Okay good, thanks :)

PR #437, opened 2020-01-31
Glopez/scenario runner

glopezdiest:
#### Description

3 atomic criterias have been changed, adding more functionalities:

- **CollisionTest** now records the location of the collision and if another one is detected close-by, it is ignored (fixes multiple micro-collsiions with static objects). If the vehicle moves away from the collision location, that location is forgotten, and hits on the same location will count once again

- **WrongLaneTest** has been changed to the update method as it now records the amount of distance traversed through the wrong lane. While in a junction, no wrong lane exists and the distance isn't increased.

- **SidewalkTest** has been redesigned to also detect when a vehicle goes out of the lane but not into a sidewalk. As with WrongLaneTest, it now also records the meters traversed. Additionally, now correctly detects that the vehicle isn't at a driving lane when inside junctions.

  * **Python version(s):** 3.6
  * **CARLA version:** 0.9.7

#### Possible Drawbacks

Sidewalks and junctions might not be 100% reliable.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/437)


glopezdiest:
Found an error, working on a fix

glopezdiest:
Fixed the bugs

fabianoboril:
Please run srunner/util/code_check_and_formatting.sh and fix the reported pylint issues (or see Travis directly).

fabianoboril:
Hi,
please rebase your branch to current master. This will resolve the remaining Travis issues (hopefully).

fabianoboril:
Why is this PR for stable? It should be for master, or not??

glopezdiest:
I talked to german and he told me it would be best if the changes were added to both master and stable

fabianoboril:
I will make a cherry-pick to the other branch. But first we should have it in master. Could you please change the PR to be on master? (You can just edit this PR).

PR #456, opened 2020-02-13
Glopez/new scenarios7 10 route

glopezdiest:
#### Description

Changed how scenarios 7 to 10 are triggered, making them more visible, instead of being based on the TrafficLightManipulator's randomness, which has a very "hidden" behavior.

- **TrafficLightManipulator** has been changed to be triggered by the new scenarios, instead of being always active in the background:

    **1.** Removed its random components
**2.** Now ends when the ego_vehicle has left the intersection
**3.** Needs a config.direction variable to decide which configuration to choose. **EDIT:** changed the config.direction variable to the subtype of the scenario. Now has an a additional check for intersections with 3 arms.

- **get_scenario_type()** at route parser has been remade to return a tag, instead of an index number. This was done to avoid creating several versions of scenario7, as it was dependent on both the route, and the TrafficLightManipulator configuration. As a consequence:

  **1.** Scenarios 7 to 9 are triggered by the same scenario SignalJunctionCrossingRoute,
**2.** A new Scenario4 has been created, VehicleTurningRoute, a copy of its non-route but with an additional parameter, "direction", which decides if the route goes left or right.

- **SignalJunctionCrossingRoute** is now responsible of triggering scenarios 7, 8 and 9. Its behavior is divided in two parts, the TrafficLightManipulator, and a DriveDistance.
- **NoSignalJunctionCrossingRoute** is the scenario that triggers Scenario10, and it ends when the ego_vehicle has left the intersection (uses new atomic trigger condition **WaitEndIntersection**)

- **annotate_trafficlight_in_group()** at carla data provider as had minor modifications:

  **1.** The "left" and "right" were switched with each other, as they were pointing to the wrong traffic_lights
**2.** Filled the gap between 300º and 340º, which, due to the else, was previously being detected as "right". 

  * **Python version(s):** 3.6
  * **CARLA version:** 0.9.7

#### Possible Drawbacks

The removal of the TrafficLightManipulator's global behavior might affect intersections where none of the scenarios 7 to 9 are triggered.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/456)


glopezdiest:
Any idea why I'm having import py_trees error? Why there and not in the other files?

fabianoboril:
> Any idea why I'm having import py_trees error? Why there and not in the other files?

You mean in Codacy? Ignore it for the moment, we have to disable this warning, as codacy currently does not use py_trees in its Pythonpath.

glopezdiest:
Travis passed! Found a problem though, don't merge it yet

glopezdiest:
Okay, all ready! And sorry for all the amount of messages you must have received...

PR #465, opened 2020-02-25
Initial support for OpenSCENARIO Catalogs

### LINKED ISSUE #459: OpenSCENARIO Catalog support
Has there been any work to support catalogs in OpenSCENARIO? They would be very useful when making a bunch of similar scenarios.

A use case would be when creating a new entity one could simply having a catalog reference rather than re-defining the same vehicle in multiple files.

I don't see any other requests for it, but I wanted to check before beginning to implement a solution.

amstrudy:
#### Description

Fixes #459 

These changes allow for the use of the Vehicle, Pedestrian, Environment, MiscObject, and Maneuver OpenSCENARIO Catalogs. By including a catalog file, one can simply reference one of its entries in multiple scenarios. 

This is done by reading in all of the catalog entries into a dictionary that resides in `OpenScenarioConfig`. In the places where Vehicles/Pedestrians/MiscObjects/Environments are read from the scenario file, the catalogs are simply used instead. Maneuvers were a bit trickier but again if a `CatalogReference` for a maneuver was present, then the maneuver is pulled from the dictionary and used.

Additionally, I have created example catalogs (they reside in the `catalogs` directory in the examples folder) and `CatalogExample.xosc` which shows how to use the catalogs in a scenario. The user must make sure the path variable at the top of the scenario file points to the `catalogs` directory.

  * **Python version(s):** 3.6.9
  * **CARLA version:** 0.9.6

#### Possible Drawbacks

I am not 100% sure if the changes I made regarding the `get_xml_path` function are correct, but they didn't seem to cause any problems for me.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/465)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@ results*.txt
 *.log
 .project
 .pydevproject
+*.png
+*.svg
+*.dot
--- a/Docs/CHANGELOG.md
+++ b/Docs/CHANGELOG.md
@@ -36,6 +36,7 @@
     - WaypointFollower atomic can handle pedestrians
     - Extensions in WaypointFollower atomic for consecutive WaypointFollowers (one WF cancels the previous one)
 * Extended OpenScenario support:
+    - Added initial support for Catalogs (Vehicle, Pedestrian, Environment, Maneuver, and and MiscObject types only)
     - Added support for UserDefinedActions (e.g. to run additional scripts)
     - Added init speed behavior for vehicles
     - Added support for relative velocities
--- a/Docs/CHANGELOG.md
+++ b/Docs/CHANGELOG.md
@@ -10,6 +10,8 @@
 * Added example scenario for lane change
 * Added cut-in example scenario
 * New Features:
+    - Added new OutsideRouteLanesTest atomic criter that encompasses both SidewalkTest and WrongLaneTest, returning
+    the percentage of route that has been traveled outside the lane.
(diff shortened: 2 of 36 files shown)


fabianoboril:
Hi,
thank you for this awesome PR!
I have only few code comments, but still need to test the review. Could you please update to latest master?

Regarding the xml_path question: This is currently used to have unique IDs for the one-shot behaviors required for py_trees (to avoid repeated execution of the same behavior).

fabianoboril:
Hi,
no worries with the commits. It is good that you keep on working on this topic. This is very much appreciated.
Regarding the issues you mentioned: Can you check if the PR #474 fixes your issues? Then we can merge it.

amstrudy:
Hi @fabianoboril,
Yes, #474 does fix the problems I was having with the weather. I just pushed one more update that merges those changes with mine.

fabianoboril:
> Hi @fabianoboril,
> Yes, #474 does fix the problems I was having with the weather. I just pushed one more update that merges those changes with mine.

Awesome. Can you update a last time to the newest master. I will then review your PR today!

amstrudy:
@fabianoboril Updated and should be good to go!

fabianoboril:
I have only some minor comments left. Testing was fine (except the absolute path for the catalog; would be good to use a relative one). Once the minor things are in, we can merge this awesome PR. Thank you for your work!

PR #475, opened 2020-02-29
Added watchdog to ScenarioManager to handle timeouts and CARLA crashes

fabianoboril:
A watchdog is added to the ScenarioManager to detect and handle internal
timeouts as well as CARLA/Unreal crashes. Upon these a KeyboardInterrupt
is raised to trigger a clean shutdown of ScenarioRunner.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/475)

### DIFF
--- a/Docs/CHANGELOG.md
+++ b/Docs/CHANGELOG.md
@@ -7,12 +7,18 @@
 * [CARLA ScenarioRunner 0.9.2](#carla-scenariorunner-092)
 
 ## Latest Changes
+### :rocket: New Features
+* Added "--timeout" command line parameter to set a user-defined timeout value
+* OpenSCENARIO support:
+    - Add initial support for Catalogs (Vehicle, Pedestrian, Environment, Maneuver, and and MiscObject types only)
 ### :bug: Bug Fixes
 * Fixed #471: Handling of weather parameter (cloudyness -> cloudiness adaption)
 * Fixed #472: Spawning issue of pedestrians in OpenSCENARIO
 * Fixed #374: Usage of evaluation critieria with multiple ego vehicles in OpenSCENARIO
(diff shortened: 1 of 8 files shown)


glopezdiest:
The print("watchdog exception") is working, and I can see it through the terminal but after that, the simulation just freezes and it doesn't end.

Tested this by starting a route scenario and then manually closing Unreal. Then the print was shown, but after restarting Unreal it could be seen that the previous simulation was still running

fabianoboril:
> The print("watchdog exception") is working, and I can see it through the terminal but after that, the simulation just freezes and it doesn't end.
> 
> Tested this by starting a route scenario and then manually closing Unreal. Then the print was shown, but after restarting Unreal it could be seen that the previous simulation was still running

This is due to the required cleanup. SR tries to destroy all actors on cleanup (required for proper shutdown). An this causes timeouts. Reduce the client timeout from 30s to 1s and you will see the clean shutdown. We can discuss to reduce the timeout to 1-2 seconds if we detect such an issue, however this may have side effects in a server environment.

glopezdiest:
> This is due to the required cleanup. SR tries to destroy all actors on cleanup (required for proper shutdown). An this causes timeouts. Reduce the client timeout from 30s to 1s and you will see the clean shutdown. We can discuss to reduce the timeout to 1-2 seconds if we detect such an issue, however this may have side effects in a server environment.
>
Then the default value shouldn't be 1 sec, but higher than 30, and if someone changes it, that's up to them

fabianoboril:
> > This is due to the required cleanup. SR tries to destroy all actors on cleanup (required for proper shutdown). An this causes timeouts. Reduce the client timeout from 30s to 1s and you will see the clean shutdown. We can discuss to reduce the timeout to 1-2 seconds if we detect such an issue, however this may have side effects in a server environment.
> 
> Then the default value shouldn't be 1 sec, but higher than 30, and if someone changes it, that's up to them

Done

PR #478, opened 2020-03-02
Glopez/new s7 to 10 behaviour

glopezdiest:
#### Description

**Note:** The commits here are a bit odd as this branch was used for different testings. I'd suggest ignoring the commits, checking the whole branch and squashing

Remade **TrafficLightManipulator**. Now the sequence is as follows:
1. When triggered, sets all the lights to red except the "ego lane", which is set to green. This is done to make the possible queue of vehicle start moving
2. When close to the intersection, the traffic light is set to yellow and then to red. Another branch of the junction (depending on the chosen scenario) is set to green to let vehicle enter the junction.
3. After some time, the ego lane is set to green. There is a waiting period to allow the vehicle to enter the junction, increasing the odds of collisions.
4. Resets the junctions

**Important:** The distance at step 2 and waiting time at step 3 are experimental and should be checked in detail (depends on the agent, speed of the vehicles in the queue, stopping at red lights algortihm...)

Minor fixes:
- **annotate_trafficlight_in_group** at CarlaDataProvider now uses the lane waypoint yaw instead of the the yaw of the traffic lights' pole, which is less reliable.
- Fixed some angle related bugs causing scenarios to not spawn (**check_trigger_position** and **match_world_location_to_route**)
- When checking if a scenario is valid (**get scenario_type**), now ignores waypoint with road options equal to lane changes.

**Question**: I added **get_trafficlight_trigger_location** function at CarlaDataProvider. It best fits as scenario_helper.py but this creates an import error due to circular dependencies. Is there a way to prevent this?

  * **Python version(s):** 3.6
  * **CARLA version:** 0.9.7.3

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/478)


glopezdiest:
> This is more maintenance than new feature, isn't it?
Hmm.. the algorithm has been completely changed, but nothing new has been added. Is that considered maintenance?

fabianoboril:
The PR makes use of #475. We should first merge #475 or remove the dependency.

glopezdiest:
Yeah I used this branch to test it. Quite sure I removed all the #475 changes though.

PR #482, opened 2020-03-06
New triggering condition for scenarios inside a route

glopezdiest:
#### Description

This gonna be a long description, and I'm sorry for that, but it is a very core part so precautions must be taken.

**Origin:** This new scenario trigger condition is motivated by the fact that when two scenarios are at adjacent lanes, they will both be triggered at the same time. this was "patched" by adding a * 2 at the route_parser.

This **new triggering condition** is meant to take into account the **route distance instead of the euclidean distance**, to avoid both scenarios from being triggered at the same time. This new type of distance measurement might also open up other areas.

**Previous Behavior**:
BasicScenario added the trigger condition **InTriggerDistancetoLocationAlongRoute** 
RouteScenario added the **oneshot behavior** (Selector between a blackboard variable and the scenario behavior), to avoid repetitions of the scenarios.

New_trigger_condition_1

**New Behavior**:

The scenarios are now a **Sequence** of:
- **WaitForBlackboardVariable**: AtomicTriggerCondition that waits for a specific variable to be True.
- **Scenario Behavior**:
- Blackboard variable reset:

This blackboard variables are managed via a parallel behavior (**ScenarioTriggerer**), which will set them to True if the ego vehicle is close by to the trigger points of the scenarios. The oneshot is also inside this new behavior (a repeat argument has been added for possible future utilities, but it is currently deactivated).

All of this is only activated if the scenario's configuration has the "route_var_name" attribute different from None, if not, the old InTriggerDistanceToLocationAlongRoute + oneshot combo is used.

New_trigger1

Also, **TrafficJamChecker** has been removed from BackgroundAcitivity, as TM now handles that 

  * **Python version(s):** 3.6
  * **CARLA version:** 0.9.7.4

#### Possible Drawbacks

Might break depending on how users are using the RouteScenario class. (Although old trigger is exactly the same). 
Scenarios outside routes shouldn't be affected, but they might

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/482)

### DIFF
--- a/Docs/CHANGELOG.md
+++ b/Docs/CHANGELOG.md
@@ -10,8 +10,12 @@
 ### :rocket: New Features
 * OpenSCENARIO support:
     - Added support for position with Lane information  (roadId and laneId)
+    - Added support for TimeOfDay tag
 * Scenario updates:
     - Scenarios that are part of RouteScenario have had their triggering condition modified. This will only activate when a certain parameter is set, and if not, the old trigger condition will still be applied.
+
+### :bug: Bug Fixes
+* Fixed spawning bugs for scenario DynamicObjectCrossing when it was part of a route
 
 ## CARLA ScenarioRunner 0.9.8
 ### :rocket: New Features
--- a/Docs/CHANGELOG.md
+++ b/Docs/CHANGELOG.md
@@ -15,7 +15,8 @@
     - Scenarios that are part of RouteScenario have had their triggering condition modified. This will only activate when a certain parameter is set, and if not, the old trigger condition will still be applied.
 
 ### :bug: Bug Fixes
-* Fixed spawning bugs for scenario DynamicObjectCrossing when it was part of a route
+* Fixed spawning bugs for scenario DynamicObjectCrossing when it is part of a route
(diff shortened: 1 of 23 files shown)


fabianoboril:
Hi,
I did some testing with the changes in BasicScenario. It works well with OpenSCENARIO, but breaks the old python-based scenarios. The problem is that for the latter, the start condition ensured that the ego vehicle is in the scene, and is _moving_. Now, as soon as the ego vehicle is available, all actors start doing whatever they should do. Hence, the behavior might be over before the ego_vehicle actually arrives at the location where something should happen. Maybe wen can keep the start condition, as long as there is no route present?

glopezdiest:
Yes, you are indeed right.

When I checked the code I understood that it was only used for routes, that's why I just commented everything, but there is the InTimeToArrivalToLocation condition, which handles the "wait until it is moving" problem you saw. I just removed the route condition, and now returns None if they are part of a route.

It might be good to put the new trigger condition here but as it also needs the ScenarioTriggerer, I haven't found a clean way of doing it.

PR #508, opened 2020-04-08
Routes now have their weather as a possible parameter

glopezdiest:
#### Description

The way routes compute the weather of the simulation has been changed to allow the user to decide the weather of each route.

To do so, the _weather_ attribute has to be added to the .xml file of the route, as a child of the _route_ attribute, along with the parameter values:
Weather

If none is chosen by the user, the default is a sunny midday (instead of the map's default one)

  * **Python version(s):** 3.6
  * **CARLA version:** 0.9.8

#### Possible Drawbacks

None?

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/508)

### DIFF
--- a/srunner/challenge/routes_training.xml
+++ b/srunner/challenge/routes_training.xml
@@ -1,17 +1,17 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <routes>
    <route id="0" map="Town01">
-		<weather 
-			cloudiness="0"
-			precipitation="0"
-			precipitation_deposits="0"
-			wind_intensity="0"
-			sun_azimuth_angle="0"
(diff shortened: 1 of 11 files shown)


srunner/challenge/routes_training.xml
```diff
@@ -1,6 +1,17 @@
...
+			cloudiness="0"
+			precipitation="0"
+			precipitation_deposits="0"
+			wind_intensity="0"
+			sun_azimuth_angle="0"
+			sun_altitude_angle="70"
+			fog_density="0"
+			fog_distance="0"
+			wetness="0"
+		/>
```
germanros1987, line 14: Fix indentation.

PR #521, opened 2020-05-02
Added initial speed support for pedestrians in OSC

fabianoboril:
#### Description

Added initial speed support for pedestrians in OSC

  * **Python version(s):** 2.7
  * **CARLA version:** 0.9.9

#### Possible Drawbacks
None

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/521)

### DIFF
--- a/Docs/CHANGELOG.md
+++ b/Docs/CHANGELOG.md
@@ -9,8 +9,12 @@
 * [CARLA ScenarioRunner 0.9.2](#carla-scenariorunner-092)
 
 ## Latest Changes
+### :rocket: New Features
+* OpenSCENARIO support:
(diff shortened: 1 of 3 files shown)


glopezdiest:
So this doesn't make pedestrians start with an initial speed, correct? That will be implemented in the future, that's what you mean by "initial support"

fabianoboril:
> So this doesn't make pedestrians start with an initial speed, correct? That will be implemented in the future, that's what you mean by "initial support"

After this review is merged, you can set an initial speed of a pedestrian.

glopezdiest:
Okay, I was confused by the walker controller as I haven't used it yet.

PR #522, opened 2020-05-04
Added manual control log and playback

glopezdiest:
#### Description

The manual control can now log the actions the ego vehicle has done to a .json file. These can also be read by the manual control to execute the same behavior.

Two new arguments have been implemented to the manual control, _log_ and _playback_, used for the functionality above mentioned. Additionally, ScenarioRunner now has a _sync_ argument, that allows to put SR in sync mode even outside routes.

When playback is active, checks have also been added to ensure the setting and ego vehicle are the same from the logged simulation. If this isn't the case, an exception is thrown. 

A socket has been added between manual_control and SR to ensure that the main loop at scenario_manager.py waits for the manual control to parse the control. Without it, a deterministic behavior cannot be ensured

CarlaDataProvider's sync flag is now updated on_carla_tick()

  * **Python version(s):** 3.7
  * **CARLA version:** 0.9.9

#### Possible Drawbacks

Sockets need to be revised as it has been a new topic for me

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/522)


glopezdiest:
@fabianoboril Hey, I'm gonna answer from here as all of your questions are related.

The socket at SM is meant to ensure determinism when running the manual control. This is by default disabled, and is only activated when wanting to log or play back certain actions. That's why the socket is inactive when args.scenario (self._scenario_mode) is not used (as routes already tick the agent from SM, so determinism is ensured). **Question: can OSC use manual_control?**

Regarding other agents, they will work as normal, and can completely ignore the socket BUT, if they want to be deterministic, they now have a socket they can connect to.

self._continue is used as the communication between manual_control and SM. If the socket is connected, it is set to False after each tick, and set to true when SM receives a message from the socket (aka manual_control has finished parsing the control).

fabianoboril:
As discussed, we will put this on a stale branch and start cherry-picking. Therefore, I suggest to close this PR.

PR #608, opened 2020-08-03
[WIP] Expose traffic manager port.

ICGog:
I'm creating this WIP PR to discuss what's the best way to set the traffic manager port so that one can run multiple simulators and scenarios on a single machine. 

To make this change, I've started from the points where the code invokes `get_trafficmanager()` or `SetAutoPilot` and propagated up the traffic manager port. However, it looks like this requires the port to be passed to BasicScenario, and hence to the constructors of all the different scenarios. What do you think, is this the right approach or is there a more contained way to make this change?

#### Description

Fixes #  

  * **Platform(s):** ...
  * **Python version(s):** ...
  * **CARLA version:** ...

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/608)

### DIFF
--- a/Docs/CHANGELOG.md
+++ b/Docs/CHANGELOG.md
@@ -82,7 +82,8 @@
 * Fixed bug causing NoSignalJunctionCrossing to not output the results of the scenario
 * Fixed bug causing SyncArrival to fail when the actor was destroyed after the behavior ended
 * Fixed bug with ending roads near stop signals to break the simulation
-
+### :ghost: Maintenance
+* Exposed traffic manager port flag to enable the execution of multiple scenarios on a single machine.
 
 ## CARLA ScenarioRunner 0.9.9
 ### :rocket: New Features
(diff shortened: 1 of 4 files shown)


fabianoboril:
Hi,

first of all, thank you very much for your contribution, which is very welcome :)

Regarding your proposal, I would recommend some adaptations to simplify the whole thing.
- Instead of providing the TM port to each routine, I would recommend to make the CarlaDataProvider aware of the TM port the current SR instance should use. For example
`@staticmethod
 def set_traffic_manager_port(new_TM_port):
   CarlaDataProvider._traffic_manager_port = new_TM_port
`
- Call this new function from scenario_runner.py to set the TM port. 
Like this you avoid providing the port through all different function/class interfaces.

ICGog:
@fabianoboril thanks for the guidance. I've addressed your comments.

PR #646, opened 2020-09-30
Not allowing negative target speeds

### LINKED ISSUE #399: In OpenSCENARIO defined command driving backward should be supported.
@fabianoboril Please, help me for that.

I want to define negative velocity in the OpenSCENARIO scenario as in line 124: https://gist.github.com/jS5t3r/c1e2543ede9275b0c735b5100d3ef7c0

But the car does not move backwards. Probably, it must be programmed that by a negative velocity 
the car sets its reverse gear. For example, by a negative velocity, a pedestrian can walk backwards.

seowwj:
#### Description

Current situation: Vehicle does not move when negative target speeds are used.

Throws an exception for vehicle controllers `npc_vehicle_control` and `pedestrian_control` when it encounters a negative target speed.
Using `simple_vehicle_control` or `vehicle_longitudinal_control` allows for negative speeds, and the vehicle will reverse as expected.

Fixes #399  

  * **Python version(s):** 2.7
  * **CARLA version:** 0.9.9.4

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/646)


fabianoboril (changes requested):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/646)* status: all files reviewed, 1 unresolved discussion (waiting on @seowwj)

___
*[srunner/scenariomanager/actorcontrols/vehicle_longitudinal_control.py, line 58 at r1](https://reviewable.io/reviews/carla-simulator/scenario_runner/646#-MIUTD6B2PaOPHyWny1Y:-MIUTD6C4FfMTdMdVgTt:b7w6ybn) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/8b4cdd49e76057b0fd019f49a94275f8ea1ade01/srunner/scenariomanager/actorcontrols/vehicle_longitudinal_control.py#L58)):*
> ```Python
>         # If target speed is negavite, raise an exception
>         if self._target_speed < 0:
>             raise NotImplementedError("Negative target speeds are not yet supported")
> ```

I think we can make this work with negative speed. Before running self._actor.apply_control(control) check the sign and go to reverse? And the if speed < target_speed should then use absolute values.

seowwj:
Thanks for the idea, I'm using a slightly different approach, which is to directly check the sign before applying the throttle.

fabianoboril (changes requested):
> Reviewed 2 of 3 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/646)* status: all files reviewed, 2 unresolved discussions (waiting on @seowwj)

___
*[srunner/scenariomanager/actorcontrols/pedestrian_control.py, line 61 at r3](https://reviewable.io/reviews/carla-simulator/scenario_runner/646#-MIVI1p26OQmMkJQ_ibZ:-MIVI1p26OQmMkJQ_ib_:b-buv9kb) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/6f73144cde6d1b4aa60b1a64b9bc36d6e4059d1e/srunner/scenariomanager/actorcontrols/pedestrian_control.py#L61)):*
> ```Python
> 
>         # If target speed is negavite, raise an exception
>         if control.speed < 0:
> ```

better make the check on target_speed!
___
*[srunner/scenariomanager/actorcontrols/vehicle_longitudinal_control.py, line 60 at r3](https://reviewable.io/reviews/carla-simulator/scenario_runner/646#-MIVIEeQ6YpGziDBwlZ6:-MIVIEeQ6YpGziDBwlZ7:b-313hte) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/6f73144cde6d1b4aa60b1a64b9bc36d6e4059d1e/srunner/scenariomanager/actorcontrols/vehicle_longitudinal_control.py#L60)):*
> ```Python
>             control.throttle = 1.0
>         elif current_speed > self._target_speed and self._target_speed < 0:
>             control.reverse = True
> ```

If the car goes first reverse, and then should move forward, I think it will never switch into forward gear. Please update this part.

fabianoboril (approved):
> Reviewed 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/646)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #658, opened 2020-10-09
Move carla.Transform to atomics for OSC Waypoints

seowwj:
#### Description

Moving transformation of OSC Waypoints to Carla positions in atomics, and added an exception when no waypoints are defined in scenario file.

Fixes #  

  * **Python version(s):** 2.7
  * **CARLA version:** 0.9.9.4

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/658)


fabianoboril (changes requested):
> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/658)* status: all files reviewed, 2 unresolved discussions (waiting on @seowwj)

___
*[srunner/scenariomanager/scenarioatomics/atomic_behaviors.py, line 544 at r2](https://reviewable.io/reviews/carla-simulator/scenario_runner/658#-MJgRlGOE7Wb7f4E9YOr:-MJgRlGOE7Wb7f4E9YOs:b-z5e7mt) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/080927cdb0e15a6e0e9cf34cc1595fc7bb536896/srunner/scenariomanager/scenarioatomics/atomic_behaviors.py#L544)):*
> ```Python
>     Args:
>         actor (carla.Actor): Controlled actor.
>         waypoints (List of carla.Transform): List of waypoints (CARLA transforms).
> ```

This should be updated, as the data has changed.
___
*[srunner/scenariomanager/scenarioatomics/atomic_behaviors.py, line 591 at r2](https://reviewable.io/reviews/carla-simulator/scenario_runner/658#-MJgRoeG9bGYfIFhfuce:-MJgRoeG9bGYfIFhfucf:b-5ig3lj) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/080927cdb0e15a6e0e9cf34cc1595fc7bb536896/srunner/scenariomanager/scenarioatomics/atomic_behaviors.py#L591)):*
> ```Python
>             carla_waypoints.append(carla_transforms)
>         self._waypoints = carla_waypoints
>         
> ```

Please run the code_format tool in srunner/utils/

seowwj:
Updated with new commits based on feedback

glopezdiest:
I don't think we should implement this change. For me, we should try to have as little atomics that have OSC inputs as possible, doing all the OSC calculus before sending the information to them.

For *ChangeActorWaypoints*, while this, along with other atomics, have been thought to be used by OSC controllers, the way they are implemented right now allow users to use them even via scenarios created with python. I won't say for sure that they are compatible with those scenarios, as we've never done extensive testing there, but with this change, that possible compatibility is broken for sure.

fabianoboril:
> I don't think we should implement this change. For me, we should try to have as little atomics that have OSC inputs as possible, doing all the OSC calculus before sending the information to them.
> 
> For _ChangeActorWaypoints_, while this, along with other atomics, have been thought to be used by OSC controllers, the way they are implemented right now allow users to use them even via scenarios created with python. I won't say for sure that they are compatible with those scenarios, as we've never done extensive testing there, but with this change, that possible compatibility is broken for sure.

Unfortunately, I disagree here. The problem with routes is, that route points can be relative positions. Hence, these have to be extracted when the maneuver starts (in theory, we even have to update the positions while the maneuver is ongoing...). As a result, we cannot convert the XML entries to CARLA position before constructing the tree.

glopezdiest:
> The problem with routes is, that route points can be relative positions.

Ah yeah, I stand corrected then, I forgot about relative positions

fabianoboril (approved):
> Reviewed 2 of 2 files at r3, 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/658)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #665, opened 2020-10-16
TimeToCollisionCondition and TimeToHeadwayCondition tags are improvised.

Sumza:
#### Description

`InTimeToArrivalToVehicle` class has been modified to support `TimeToCollisionCondition` tag for dynamic actors.

The formula used for  TimeToCollision calculation in previous Implementation is : 

```math
  Relative Distance = Distance between centre location of two actors.
  TimeToCollision = (2 * Relative Distance ) /  relative velocity of two actors
```

The formula used for  TimeToCollision calculation in current implementation is : 
```math
Relative Distance = Distance between centre location of two actors.
TimeToCollision = (Relative Distance - half of length of actor1 -  half of length of actor2 ) / relative velocity of two actors.
```
The modified calculation is done as in the below image :
ttc

The class `InTimeToArrivalToVehicle` is updated such that `TimeToHeadwayCondition` tag behaviour is not effected.

  * **Platform(s):**  Ubuntu 18.04
  * **Python version(s):**  Python 3.6
  * **Unreal Engine version(s):**  
  * **CARLA version:**  CARLA_0.9.10
#### Possible Drawbacks : None

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/665)

### DIFF
--- a/Docs/CHANGELOG.md
+++ b/Docs/CHANGELOG.md
@@ -110,6 +110,7 @@
 * Fixed handling of OSC LanePosition (#625)
 * Fixed bug causing the route repetitions to spawn different background activity
 * Fixed bug causing the rotate_point function inside RunningRedLightTest to not function properly.
+* Fixed bug causing TimeToCollisionCondition which was not working for dynamic actors.
 ### :ghost: Maintenance
 * Exposed traffic manager port flag to enable the execution of multiple scenarios on a single machine.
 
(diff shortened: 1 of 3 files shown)


glopezdiest:
Hey @Sumza, thanks for the PR. One question regarding the formula you used. According to the OSC docs, the distance calculated is the one between reference points, and only if the *freespace* attribute is active that the bounding boxes are taken into account. However, your new implementation formula always takes them into account. Is my understanding correct?

Sumza:
Hi @glopezdiest , Thanks for the feedback. Above point is noted and i will be creating a new class for ```TimeToCollisionCondition ``` and considering the freespace attribute. 
I will be creating a new PR for the above mentioned points.

glopezdiest:
Hey, I just saw your other PR (#666) so can we consider this one as closed?

Sumza:
Hi, Yes you can consider this as closed and consider the new PR for review. Thankyou!

PR #666, opened 2020-10-19
Defined a new class for TimeToCollisionCondition along with freespace implementation

Sumza:
Description

A new class ```InTimeToCollisionToVehicle``` is added to support ```TimeToCollisionCondition``` tag for dynamic actors.

The formula used for TimeToCollision calculation in previous Implementation is :
```
  Relative Distance = Distance between centre location of two actors.
  TimeToCollision = (2 * Relative Distance ) /  relative velocity of two actors
```
The formula used for TimeToCollision calculation in current implementation is :

```
Relative Distance = Distance between centre location of two actors.

When "freespace" is true:
TimeToCollision = (Relative Distance - half of length of actor1 -  half of length of actor2 ) / relative velocity of two actors.

When "freespace" is false:
TimeToCollision = Relative Distance  / relative velocity of two actors.
```
The modified calculation is done as in the below image :

   ```When "freespace" is true:```
   ttc

  ```When "freespace" is false:```
   ttc1

Where has this been tested?
Platform(s): Ubuntu 18.04
Python version(s): Python 3.6
Unreal Engine version(s):
CARLA version: CARLA_0.9.10

Possible Drawbacks : None

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/666)

### DIFF
--- a/Docs/CHANGELOG.md
+++ b/Docs/CHANGELOG.md
@@ -121,6 +121,7 @@
     - Added support to use a non-CARLA OpenDRIVE map (instead of CARLA towns)
     - Added support for TimeOfDay tag
     - Added support for scenarios with no actors
+    - Added support for TimeToCollisionCondition with freespace.
 * Scenario updates:
     - Scenarios that are part of RouteScenario have had their triggering condition modified. This will only activate when a certain parameter is set, and if not, the old trigger condition will still be applied.
 * Atomics:
--- a/Docs/CHANGELOG.md
+++ b/Docs/CHANGELOG.md
@@ -122,6 +122,7 @@
     - Added support for TimeOfDay tag
     - Added support for scenarios with no actors
     - Added support for TimeToCollisionCondition with freespace.
+    - Added support for TimeHeadwayCondition with freespace.
 * Scenario updates:
     - Scenarios that are part of RouteScenario have had their triggering condition modified. This will only activate when a certain parameter is set, and if not, the old trigger condition will still be applied.
 * Atomics:
--- a/srunner/tools/openscenario_parser.py
+++ b/srunner/tools/openscenario_parser.py
@@ -68,7 +68,6 @@ from srunner.tools.py_trees_port import oneshot_behavior
 
 
 class OpenScenarioParser(object):
-
     """
     Pure static class providing conversions from OpenSCENARIO elements to ScenarioRunner elements
     """
@@ -614,7 +613,7 @@ class OpenScenarioParser(object):
 
         if condition.find('ByEntityCondition') is not None:
 
-            trigger_actor = None    # A-priori validation ensures that this will be not None
+            trigger_actor = None  # A-priori validation ensures that this will be not None
             triggered_actor = None
 
             for triggering_entities in condition.find('ByEntityCondition').iter('TriggeringEntities'):
--- a/srunner/tools/openscenario_parser.py
+++ b/srunner/tools/openscenario_parser.py
@@ -54,7 +54,6 @@ from srunner.scenariomanager.scenarioatomics.atomic_trigger_conditions import (I
                                                                                InTriggerDistanceToOSCPosition,
                                                                                InTimeToArrivalToOSCPosition,
                                                                                InTimeToArrivalToVehicle,
-                                                                               InTimeToCollisionToVehicle,
                                                                                DriveDistance,
                                                                                StandStill,
                                                                                OSCStartEndCondition,
@@ -678,10 +677,7 @@ class OpenScenarioParser(object):
                     condition_rule = headtime_condition.attrib.get('rule')
                     condition_operator = OpenScenarioParser.operators[condition_rule]
 
-                    condition_freespace = strtobool(headtime_condition.attrib.get('freespace', False))
-                    if condition_freespace:
-                        raise NotImplementedError(
-                            "TimeHeadwayCondition: freespace attribute is currently not implemented")
+                    condition_freespace = strtobool(headtime_condition.attrib.get('freespace'))
                     condition_along_route = strtobool(headtime_condition.attrib.get('alongRoute', False))
 
                     for actor in actor_list:
@@ -693,7 +689,7 @@ class OpenScenarioParser(object):
                             headtime_condition.attrib.get('entityRef', None)))
 
                     atomic = InTimeToArrivalToVehicle(
-                        trigger_actor, triggered_actor, condition_value,
+                        trigger_actor, triggered_actor, condition_value, condition_freespace,
                         condition_along_route, condition_operator, condition_name
                     )
 
@@ -723,7 +719,7 @@ class OpenScenarioParser(object):
                             raise AttributeError("Cannot find actor '{}' for condition".format(
                                 entity_ref_.attrib.get('entityRef', None)))
 
-                        atomic = InTimeToCollisionToVehicle(
+                        atomic = InTimeToArrivalToVehicle(
                             trigger_actor, triggered_actor, condition_value, condition_freespace,
                             condition_along_route, condition_operator, condition_name)
                 elif entity_condition.find('AccelerationCondition') is not None:
--- a/srunner/tools/openscenario_parser.py
+++ b/srunner/tools/openscenario_parser.py
@@ -20,49 +20,57 @@ import carla
 
 from srunner.scenariomanager.carla_data_provider import CarlaDataProvider
 from srunner.scenariomanager.weather_sim import Weather
-from srunner.scenariomanager.scenarioatomics.atomic_behaviors import (TrafficLightStateSetter,
-                                                                      ActorTransformSetterToOSCPosition,
-                                                                      RunScript,
-                                                                      ChangeWeather,
-                                                                      ChangeAutoPilot,
-                                                                      ChangeRoadFriction,
-                                                                      ChangeActorTargetSpeed,
-                                                                      ChangeActorControl,
-                                                                      ChangeActorWaypoints,
-                                                                      ChangeActorWaypointsToReachPosition,
-                                                                      ChangeActorLateralMotion,
-                                                                      Idle)
+from srunner.scenariomanager.scenarioatomics.atomic_behaviors import (
+    TrafficLightStateSetter,
+    ActorTransformSetterToOSCPosition,
+    RunScript,
+    ChangeWeather,
+    ChangeAutoPilot,
+    ChangeRoadFriction,
+    ChangeActorTargetSpeed,
+    ChangeActorControl,
+    ChangeActorWaypoints,
+    ChangeActorWaypointsToReachPosition,
+    ChangeActorLateralMotion,
+    Idle,
+)
+
 # pylint: disable=unused-import
 # For the following includes the pylint check is disabled, as these are accessed via globals()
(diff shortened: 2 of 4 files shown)


glopezdiest:
Thanks for the PR. I'll try to test it this afternoon and give you feedback. One question regarding the code, why are there so many blank lines deleted inside comments?

Sumza:
Sorry for the inconvenience,I had tried formatting the code through an editor which might have caused the deletion of blank lines. Please let me know if this will be an issue, will fix and commit the changes.

glopezdiest:
I'd very much appreciate it. In my opinion, comments were clearer before. In addition, the amount of changes will be reduced, helping us track the actual code changes better.

Sumza:
I have updated the code by adding blank lines in the comments as before. Please consider the recent commit.

glopezdiest:
Okay, I have three topics to discuss from your code:

First of all, you are **only focusing on** checking the distance between two **vehicles that are going straight through a lane**. What this means is that if the reference entity was perpendicular to a lane, this wouldn't correctly calculate the distance. It is because of this that we didn't implement it in the first place, but right now, I'm not sure whether we should care too much about it, as it's quite a corner case.

Secondly, there is **no need to actually create a new class**. I'd vote to add a new argument to *InTimeToArrivalToVehicle*, *freespace*, just before the comparison argument, which is by default False. With that, we avoid having duplicate atomics. The only conflict would be with *TimeHeadwayCondition*, but that one is easily resolved (which by the way, also needs the *freespace* implemented... )

Lastly, regarding the formula itself, you can't just get the abs value of the velocities, as **if you are slower than the reference, you'll never catch it**. Also, from our side, we have a *2* at ``` time_to_arrival = 2 * distance / (current_velocity - other_velocity) ```. However, I don't understand why it is there (all the physics I know say to me that it shouldn't be there).

I'll think about the first topic and report back

Sumza:
Hi, According to you second and third point, below changes are made in the code :

1) The class "InTimeToCollisionToVehicle" is removed and a single class "InTimeToArrivalToVehicle" is maintained for both 
    "TimeToCollisionCondition" and "TimeToHeadwayCondition". 
2) attribute "freespace" is considered for both "TimeToCollisionCondition" and "TimeToHeadwayCondition" tags. 
    abs value of velocities is not considered since calculation is done only when ```current_velocity > other_velocity``` as
    per your third point. Below is the code changes done in "InTimeToArrivalToVehicle" class :
```       
  if current_velocity > other_velocity:
            if self._condition_check == "TimeToCollisionCondition" :
                if self._condition_freespace:
                    time_to_arrival = (distance - self._actor.bounding_box.extent.x -
                                       self._other_actor.bounding_box.extent.x) / (current_velocity - other_velocity)
                else:
                    time_to_arrival = distance / (current_velocity - other_velocity)

            else:
                if self._condition_freespace:
                    time_to_arrival = 2 * (distance - self._actor.bounding_box.extent.x -
                                           self._other_actor.bounding_box.extent.x)/ (current_velocity - other_velocity)
                else:
                    time_to_arrival = 2 * distance / (current_velocity - other_velocity)
```
3) Since "InTimeToArrivalToVehicle" class is also called in ```object_crash_vehicle.py``` , the below changes have been made in the file to maintain the same behaviour as before :
```
            condition_check = "Object_crash_vehicle"
            condition_freespace = "False"
            start_condition = InTimeToArrivalToVehicle(self.ego_vehicles[0],
                                                       self.other_actors[0],
                                                       self._time_to_reach,
                                                       condition_check,
                                                       condition_freespace)
```
4) For "TimeToHeadwayCondition" , 2 in ```time_to_arrival``` calculation is maintained as before in the master code, can you confirm if this needs to be maintained the same or modified?

Please provide your feedback on the above points.

glopezdiest:
Hey, you haven't updated any code so I can't tell you for sure. However:

> The class "InTimeToCollisionToVehicle" is removed and a single class "InTimeToArrivalToVehicle" is maintained for both
"TimeToCollisionCondition" and "TimeToHeadwayCondition".

This seems perfect.

> attribute "freespace" is considered for both "TimeToCollisionCondition" and "TimeToHeadwayCondition" tags.
abs value of velocities is not considered since calculation is done only when current_velocity > other_velocity as
per your third point

Also perfect.

> 2 in time_to_arrival calculation is maintained as before in the master code, can you confirm if this needs to be maintained the same or modified?

Either I communicated myself unclearly or you missunderstood what I meant. That *2* in the formula confused me. Right now I can't see any reason why it should be there, so let's remove it altogether. This means that your *condition_check* variable is unneeded as they all have the same formula. 

> Since "InTimeToArrivalToVehicle" class is also called in object_crash_vehicle.py , the below changes have been made in the file to maintain the same behaviour as before :

If you define the *freespace* argument with a default value of *False*, you don't have to modify the *object_crash_vehicle.py* plus we maintain backwards compatibility with that scenario.

Sumza:
Hi , can you please confirm me with the below points :

1) By removing 2 from the formula ,  "TimeToCollisionCondition" and "TimeToHeadwayCondition" will have the same formula 
   calculations. Is that okay ?

2) According to my understanding as we have implemented  "freespace" for both "TimeToCollisionCondition" and 
   "TimeToHeadwayCondition", the argument freespace has to be passed in function call of "TimeToArrivalToVehicle" class in  
    object_crash_vehicle.py as well. 
   Please correct me if i am wrong.

3) If "freespace" argument is maintained as false by default, is it okay to subtract the half of the lengths of 
    vehicles in timetoarrival formula ? If yes, will implement the same and commit the code.

glopezdiest:
1) Yes

2) No. Before this change, all the current calculus are getting done without taking into account the bounding boxes of the vehicles. If you modify the initialize of *InTimeToArrivalToVehicle* to be

```
def __init__(self, actor, other_actor, time, along_route=False, freespace=False,
             comparison_operator=operator.lt, name="TimeToArrival"):
```

the *object_crash_vehicle.py* call won't have to be modified. So by default, the atomic will still calculate the time without taking into account the bounding boxes, the same as it was before, and only if you pass the argument will the atomic change.

3) Your formula is already correct:

```
if self._condition_freespace:
    time_to_arrival = (distance - self._actor.bounding_box.extent.x -
                       self._other_actor.bounding_box.extent.x) / (current_velocity - other_velocity)
else:
    time_to_arrival = distance / (current_velocity - other_velocity)
```

This goes in tandem with number 2). By default *self._condition_freespace* will be False, so the *else* will be triggered, using the previous formula we were using.

Sumza:
Hi, I have a requirement for TimeToCollisionCondition to be calculated from edge to edge of two vehicles.

By making the freespace argument as false by default, the calculation will be done from centre to centre and not from edge to edge between two vehicles.

Can you let us know on how to proceed with changes to satisfy our requirement.

glopezdiest:
No, you'll be taking into account the bounding boxes, which is the same as you are doing with your code. Here
```
condition_freespace = strtobool(ttc_condition.attrib.get('freespace'))
```
you parse the freespace attribute from OSC, which is then given to the atomic *InTimeToCollisionToVehicle*
```
atomic = InTimeToCollisionToVehicle(trigger_actor, triggered_actor,
                                    condition_value, condition_freespace, condition_along_route,
                                    condition_operator, condition_name)
```
with the following initalization
```
def __init__(self, actor, other_actor, time, condition_freespace, along_route=False,
             comparison_operator=operator.lt, name="TimeToArrival"):
```

By changing *condition_freespace* to *condition_freespace=False* you make it by default equal to False. However, as you are giving that argument as an input, the *False* value will be changed to whatever you passed, which in this case is the value parsed from OSC.

Sumza:
Hi, I have updated the code as per your suggestions and pushed it. 

Please check and let me know if any more suggestions or changes to be done.

glopezdiest:
Okay, everything seems good, many thanks for the changes :slightly_smiling_face:. I'll do some testing and report back.

srunner/tools/openscenario_parser.py
```diff
@@ -702,13 +699,11 @@ def convert_condition_to_atomic(condition, actor_list):
...
-                    condition_value = ttc_condition.attrib.get('value')
+                    condition_value = float(ttc_condition.attrib.get('value'))
                     condition_target = ttc_condition.find('TimeToCollisionConditionTarget')
+                    entity_ref_ = condition_target.find('EntityRef')
 
-                    condition_freespace = strtobool(ttc_condition.attrib.get('freespace', False))
-                    if condition_freespace:
-                        raise NotImplementedError(
-                            "TimeToCollisionCondition: freespace attribute is currently not implemented")
+                    condition_freespace = strtobool(ttc_condition.attrib.get('freespace'))
```
glopezdiest, line 706: I'd keep the *False* there, as by default, if the attribute is not found, it returns None.

srunner/tools/openscenario_parser.py
```diff
@@ -677,10 +677,7 @@ def convert_condition_to_atomic(condition, actor_list):
                     condition_rule = headtime_condition.attrib.get('rule')
                     condition_operator = OpenScenarioParser.operators[condition_rule]
 
-                    condition_freespace = strtobool(headtime_condition.attrib.get('freespace', False))
-                    if condition_freespace:
-                        raise NotImplementedError(
-                            "TimeHeadwayCondition: freespace attribute is currently not implemented")
+                    condition_freespace = strtobool(headtime_condition.attrib.get('freespace'))
```
glopezdiest, line 680: I'd keep the *False* there, as by default, if the attribute is not found, it returns None.

Docs/CHANGELOG.md
```diff
@@ -121,6 +121,7 @@
     - Added support to use a non-CARLA OpenDRIVE map (instead of CARLA towns)
     - Added support for TimeOfDay tag
     - Added support for scenarios with no actors
+    - Added support for TimeToCollisionCondition with freespace.
```
glopezdiest, line 124: We are technically also supporting freespace for TimeHeadwayCondition now. Please update it. Also, the docs have to be updated. At *Docs/openscenario_support.md* search for these two conditions and remove the warning saying that freespace is not supported

glopezdiest (changes requested):
I haven't found any issue, just leaving some notes regarding documentation and minor changes. After these, we can merge the PR.

Sumza:
Thankyou, 

I have updated the changes according to your suggestions. Please check and let me know if any more changes to be done.

glopezdiest:
Perfect, everything is fine from my side. I'm updating the branch to current master and I'll merge it once it finishes

glopezdiest (approved):
> Reviewed 1 of 2 files at r3, 3 of 3 files at r5.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/666)* status: all files reviewed, 3 unresolved discussions (waiting on @Sumza)

glopezdiest (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/666)* status: :shipit: complete! all files reviewed, all discussions resolved

fabianoboril (approved):
> Reviewed 1 of 2 files at r3, 3 of 3 files at r5.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/666)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #669, opened 2020-10-19
Fixed shutdown issue in ScenarioManager causing problems in sync mode

fabianoboril:
On shutdown it could happen that sync mode remained active, causing issues on next startup

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/669)

### DIFF
--- a/Docs/CHANGELOG.md
+++ b/Docs/CHANGELOG.md
@@ -16,6 +16,7 @@
 ### :bug: Bug Fixes
 * Fixed exception when using OSC scenarios without EnvironmentAction inside Storyboard-Init
 * Fixed bug causing the TrafficManager to not be correctly updated at asynchronous simualtions
+* Fixed shutdown issue in ScenarioRunner causing to not switch to asynchronous mode
 ### :ghost: Maintenance
 * Added check to ensure OSC names (for story/act/maneuver) are unique
 
--- a/scenario_runner.py
+++ b/scenario_runner.py
@@ -170,13 +170,15 @@ class ScenarioRunner(object):
         Remove and destroy all actors
         """
         # Simulation still running and in synchronous mode?
-        if self.manager is not None and self.manager.get_running_status() \
-                and self.world is not None and self._args.sync:
-            # Reset to asynchronous mode
-            settings = self.world.get_settings()
-            settings.synchronous_mode = False
-            settings.fixed_delta_seconds = None
-            self.world.apply_settings(settings)
+        if self.world is not None and self._args.sync:
+            try:
+                # Reset to asynchronous mode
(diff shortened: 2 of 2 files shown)


glopezdiest:
Hey, have you found a way to easily reproduce it? I'd like to give it a try. Also, setting the world to async is already done also at the *cleanup* function. Is there a problem causing it to not be called?

fabianoboril:
> Hey, have you found a way to easily reproduce it? I'd like to give it a try. Also, setting the world to async is already done also at the _cleanup_ function. Is there a problem causing it to not be called?
We have seen that a timeout could cause SR to termiante, and then in the following run the world is still in sync mode. Therefore, we added that fix.

glopezdiest:
Okay, I'll give it a go

glopezdiest:
I haven't been able to reproduce the issue and I am also unable to see how this fixes the issue. As far as I can see, the *stop_scenario* function is only called during the *_signal_handler*:
```
    def _signal_handler(self, signum, frame):
        """
        Terminate scenario ticking when receiving a signal interrupt
        """
        self._shutdown_requested = True
        if self.manager:
            self.manager.stop_scenario()
            self._cleanup()
            if not self.manager.get_running_status():
                raise RuntimeError("Timeout occured during scenario execution")
```
which is just before the *_cleanup*:

```
    def _cleanup(self):
        """
        Remove and destroy all actors
        """
        # Simulation still running and in synchronous mode?
        if self.manager is not None and self.manager.get_running_status() \
                and self.world is not None and self._args.sync:
            # Reset to asynchronous mode
            settings = self.world.get_settings()
            settings.synchronous_mode = False
            settings.fixed_delta_seconds = None
            self.world.apply_settings(settings)

        self.manager.cleanup()

        CarlaDataProvider.cleanup()

        # + more cleanup code

```

I don't see the need to set the world to async at the *stop_scenario* as we are already doing right after that.

Moreover, in case the simulation is abruptly shut down, this change makes the code get stuck at the *world.get_settings()* function, which stops the cleanup from being executed. Before, this didn't happen at the cleanup because it has the watchdog check, allowing it to continue with the cleanup.

fabianoboril:
Updated the code as discussed during our call.

glopezdiest (approved):
Perfect. In short, the problem was that the watchdog could trigger for reasons other than the simulation having shut down. When this happened, the world wasn't changed back to async, due to the watchdog check. 

> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/669#-:-MK59dbT5j1dq9lKiLzK:b-tisfrt)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #685, opened 2020-11-17
Fix route scenario

dsche:
#### Description
This PR offers a fix for the case when a route scenario is run without an agent, which currently produces a runtime error since the agent is None.

  * **Platform(s):** Ubuntu 20.04.1
  * **Python version(s):** 3.8.5
  * **CARLA version:** 0.9.10

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/685)

### DIFF
--- a/Docs/CHANGELOG.md
+++ b/Docs/CHANGELOG.md
@@ -19,6 +19,7 @@
 * Fixed bug causing the TrafficManager to not be correctly updated at asynchronous simualtions
 * Fixed shutdown issue in ScenarioRunner causing to not switch to asynchronous mode
 * Fixed OSC TeleportAction within Story
+* Fixed runtime exception on RouteScenario without an agent parameter 
 ### :ghost: Maintenance
 * Added check to ensure OSC names (for story/act/maneuver) are unique
 
--- a/srunner/scenarios/route_scenario.py
+++ b/srunner/scenarios/route_scenario.py
@@ -190,7 +190,8 @@ class RouteScenario(BasicScenario):
         self.route = route
         CarlaDataProvider.set_ego_vehicle_route(convert_transform_to_location(self.route))
 
-        config.agent.set_global_plan(gps_route, self.route)
+        if config.agent is not None:
+            config.agent.set_global_plan(gps_route, self.route)
 
         # Sample the scenarios to be used for this route instance.
         self.sampled_scenarios_definitions = self._scenario_sampling(potential_scenarios_definitions)


fabianoboril (changes requested):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/685)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @dsche)

___
*[srunner/scenarios/route_scenario.py, line 193 at r1](https://reviewable.io/reviews/carla-simulator/scenario_runner/685#-MMP_dEB8koa2du4nhRt:-MMP_dEB8koa2du4nhRu:b-u0p2pw) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/de7e6d88a3f76cb6ca522d78d73a4474c5d1eec6/srunner/scenarios/route_scenario.py#L193)):*
> ```Python
>         CarlaDataProvider.set_ego_vehicle_route(convert_transform_to_location(self.route))
> 
>         if config.agent is not None:
> ```

If there is no agent, something went fundamentally wrong. I guess we should better raise an exception. I wonder why this doesn't happen...

dsche (commented):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/685)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @fabianoboril)

___
*[srunner/scenarios/route_scenario.py, line 193 at r1](https://reviewable.io/reviews/carla-simulator/scenario_runner/685#-MMP_dEB8koa2du4nhRt:-MMQT5sy6qtScIBkbn0M:b-u9yzax) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/de7e6d88a3f76cb6ca522d78d73a4474c5d1eec6/srunner/scenarios/route_scenario.py#L193)):*

If there is no agent, something went fundamentally wrong. I guess we should better raise an exception. I wonder why this doesn't happen...
</blockquote></details>

In our specific use case, this happens because our agents run independently of the scenario_runner control loop. That is, the vehicle is there, but the vehicle controlling code is not owned by the SR. Not sure if this is within SR's scope, though.

fabianoboril (approved):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/685)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #687, opened 2020-11-18
Fixed AcquirePositionAction

glopezdiest:
### Description

Fixed a bug at the last commit causing *AcquirePositionAction* of OSC to crash. The new *ChangeActorWaypoints* already has a global route planner, therefore the *ChangeActorWaypointsToReachPosition* behavior can be completely removed.

Changed some parts of the code from location to waypoint location at the *ChangeActorWaypoints*. This fixes the behaviour when the actor was initially far away from the center of the lane

  * **Python version(s):** 3.6
  * **CARLA version:** 0.9.10.1

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/687)

### DIFF
--- a/srunner/scenariomanager/scenarioatomics/atomic_behaviors.py
+++ b/srunner/scenariomanager/scenarioatomics/atomic_behaviors.py
@@ -607,8 +607,12 @@ class ChangeActorWaypoints(AtomicBehavior):
                 if i == 0:
                     mmap = CarlaDataProvider.get_map()
                     ego_location = CarlaDataProvider.get_location(self._actor)
-                    ego_waypoint = mmap.get_waypoint(ego_location).transform.location
-                    waypoint = ego_waypoint
+                    ego_waypoint = mmap.get_waypoint(ego_location)
+                    try:
+                        ego_next_wp = ego_waypoint.next(1)[0]
+                    except IndexError:
+                        ego_next_wp = ego_waypoint
+                    waypoint = ego_next_wp.transform.location
                 else:
                     waypoint = carla_route_elements[i - 1][0].location
                 waypoint_next = carla_route_elements[i][0].location
@@ -627,7 +631,7 @@ class ChangeActorWaypoints(AtomicBehavior):
                         if len(route) > 1:
                             last_heading_vec = route[-1].location - route[-2].location
                         else:
-                            last_heading_vec = route[-1].location - ego_waypoint
+                            last_heading_vec = route[-1].location - ego_next_wp.transform.location
                         last_heading = np.arctan2(last_heading_vec.y, last_heading_vec.x)
 
                         heading_delta = math.fabs(new_heading - last_heading)
--- a/srunner/scenariomanager/scenarioatomics/atomic_behaviors.py
+++ b/srunner/scenariomanager/scenarioatomics/atomic_behaviors.py
@@ -537,7 +537,6 @@ class ChangeActorWaypoints(AtomicBehavior):
     The behavior is in RUNNING state until the last waypoint is reached, or if a
     second waypoint related atomic for the same actor is triggered. These are:
     - ChangeActorWaypoints
-    - ChangeActorWaypointsToReachPosition
     - ChangeActorLateralMotion
 
     Args:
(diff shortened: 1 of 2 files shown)


fabianoboril (changes requested):
> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/687)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @fabianoboril and @glopezdiest)

___
*[srunner/scenariomanager/scenarioatomics/atomic_behaviors.py, line 610 at r1](https://reviewable.io/reviews/carla-simulator/scenario_runner/687#-MMUTi0tCKF3BkvtmpNP:-MMUTi0u-ZTp9dxwsd3U:b-9ggiy4) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/959488754dbb4595c45c4d258546ece8f9570b72/srunner/scenariomanager/scenarioatomics/atomic_behaviors.py#L610)):*
> ```Python
>                     mmap = CarlaDataProvider.get_map()
>                     ego_location = CarlaDataProvider.get_location(self._actor)
>                     ego_waypoint = mmap.get_waypoint(ego_location).transform.location
> ```

If the vehicle is on the road, can this wp be not in heading direction (i.e it is not in front of the car)? If yes, we have to add an additional check.

glopezdiest (commented):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/687)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @fabianoboril)

___
*[srunner/scenariomanager/scenarioatomics/atomic_behaviors.py, line 610 at r1](https://reviewable.io/reviews/carla-simulator/scenario_runner/687#-MMUTi0tCKF3BkvtmpNP:-MMUZax31anAbQMdemEJ:b-82iret) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/959488754dbb4595c45c4d258546ece8f9570b72/srunner/scenariomanager/scenarioatomics/atomic_behaviors.py#L610)):*

If the vehicle is on the road, can this wp be not in heading direction (i.e it is not in front of the car)? If yes, we have to add an additional check.
</blockquote></details>

This waypoint is a projection of the vehicle location into the lane, so it will always be in the center and with the heading of the lane. If the vehicle is going through a lane with the opposite direction, or is inside an intersection, the calculated waypoint might not be the one we really care about.

However, all of this is also present at the global route planner, as it only cares about the location of the vehicle. By doing this, we ensure that both the start of the route and the calculated waypoint are the same (or very similar). If we were to add additional checks, we might get a better waypoint, but when combined with the global route planner, it would make no sense.

glopezdiest (commented):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/687)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @fabianoboril)

___
*[srunner/scenariomanager/scenarioatomics/atomic_behaviors.py, line 610 at r1](https://reviewable.io/reviews/carla-simulator/scenario_runner/687#-MMUTi0tCKF3BkvtmpNP:-MMUvPdv0QUMjhwEQNZa:b-elifeu) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/959488754dbb4595c45c4d258546ece8f9570b72/srunner/scenariomanager/scenarioatomics/atomic_behaviors.py#L610)):*
<details><summary><i>Previously, glopezdiest wrote…</i></summary><blockquote>

This waypoint is a projection of the vehicle location into the lane, so it will always be in the center and with the heading of the lane. If the vehicle is going through a lane with the opposite direction, or is inside an intersection, the calculated waypoint might not be the one we really care about.

However, all of this is also present at the global route planner, as it only cares about the location of the vehicle. By doing this, we ensure that both the start of the route and the calculated waypoint are the same (or very similar). If we were to add additional checks, we might get a better waypoint, but when combined with the global route planner, it would make no sense.
</blockquote></details>

Okay, I missunderstood your comment. Changed it so that the first route point is a little bit forward than the ego vehicle location, to avoid having it behind, which might "break" the controller

fabianoboril (approved):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/687)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #709, opened 2020-12-15
Converting html tags (table, link) into markdown and fixing table styles

ebadi:
"Markdown is designed to be easy to write and to read, the content should be publishable as-is, as plain text, without looking like it has been marked up with tags or formatting instructions.", therefore HTML/JS/CSS is not recommended (while supported) to be used in markdown files. This PR convert HTML tags with their Markdown equivalents. In long term this PR enables other tools like Pandoc to work properly with these markdown files.

Html tags (table, image, link, etc.) are converted into markdown.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/709)


corkyw10 (approved):
> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/709)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #755, opened 2021-04-20
Fixed OSC parser issue on Nonetype actor

### LINKED ISSUE #753: Missing condition in ChangeActorLaneOffset (and other atomics)
Since there is a NoneType actor in the joint_actor_list in the openscenario_parser it seems it is necessary to add a additional condition in ChangeActorLaneOffset.termine(). Original code:
`if actor_dict and self._actor.id in actor_dict:`
needs:
`if actor_dict and self._actor is not None and self._actor.id in actor_dict:`

The same seems true for InTriggerDistanceToOSCPosition.update():
Original:
`if osc_transform is not None:`
to
`if osc_transform is not None and self._actor is not None:`

I am not 100% sure if this None actor is the source of the problem, but it seems to me it might cause the errors i get:

AttributeError: 'NoneType' object has no attribute 'id'
'NoneType' object has no attribute 'id'
Resetting ego-vehicle!

I think this is also a problem in other atomic behaviors/trigger that use self._actor

Could this also come from something I have to keep in mind when I build an OpenScenario file? The file is validated an does not seem to go againts OSC standards, but maybe there is something which is not allowed in Carla?

Thanks for your work!

**Desktop:**
 - OS: [Ubuntu 18.04]
 - CARLA Version [CARLA 0.9.11]
 - Python version [3.6]
 - Version [0.9.11]

fabianoboril:
Resolved exception "AttributeError: 'NoneType' object has no attribute
'attributes'" in OpenScenarioParser.

Fixes #753

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/755)

### DIFF
--- a/srunner/tools/openscenario_parser.py
+++ b/srunner/tools/openscenario_parser.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-# Copyright (c) 2019-2020 Intel Corporation
+# Copyright (c) 2019-2021 Intel Corporation
 #
 # This work is licensed under the terms of the MIT license.
 # For a copy, see <https://opensource.org/licenses/MIT>.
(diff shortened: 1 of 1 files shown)


fpasch (approved):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/755)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @fabianoboril)

PR #760, opened 2021-05-04
Extended CarlaDataProvider's spawning functions

glopezdiest:
### Description

`create_blueprint()` at the CarlaDataProvider now has a new ***safe*** argument which will perform an additional filter. This is specially useful for the creation of background activity, as it filters out 2 wheeled vehicles (bikes shouldn't be driving through the road), as well as large vehicles (which might cause problems when turning at close turns). As all spawning functions call this `create_blueprint`, they all have had the ***safe_blueprint*** argument added. This is by default False, retaining compatibility with previous versions

Another argument, ***tick***, have also been added to the spawning functions. This allows the spawning of actors without ticking the CARLA server. This is useful when running on sync mode, as in that case, the server tick must be tightly controlled. As a reference, all methods on spawned actor that haven't been ticked work the fine, but the getters will return the default value (all zeroes), as the actor wasn't alive at the previous frame. *tick* is by default True, retaining compatibility with previous versions.

Additionally, `handle_batch_actors()` has been simplified, as it was checking for impossible conditions (such as `batch` being None or `responses` being empty). Removed an exception raise at `request_new_actor()` if failing to spawn the actor, as it was inconsistent with the other two methods. It now returns None if there was a spawn error, and a fixed bug causing the autopilot to never be set.

### Where has this been tested?

  * **Python version(s):** 3.6
  * **CARLA version:** 0.9.11 (current dev)

### Possible Drawbacks

Except from some code reformatting, the rest is just extensions to the functions

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/760)

### DIFF
--- a/Docs/CHANGELOG.md
+++ b/Docs/CHANGELOG.md
@@ -22,6 +22,7 @@
 ### :bug: Bug Fixes
 * Fixed bug at the Getting Started docs which caused an import error
 * Fixed neverending lane change maneuver in OpenSCENARIO
+* Fixed bug causing the spawning of an actor with `request_new_actor` to never activate the autopilot.
 ### :ghost: Maintenance
 * Extended SimpleVehicleController (OSC) to handle traffic lights
 * Generalized visualizer attached to OSC controllers
--- a/Docs/CHANGELOG.md
+++ b/Docs/CHANGELOG.md
@@ -28,6 +28,7 @@
 * Fixed bug at the Getting Started docs which caused an import error
 * Improved the watchdog. It can now be paused, resumed and uses the same thread, instead of opening and closing new ones each frame.
 * Added `simple-watchdog-timer` library to the requirements, as it is used by the new watchdog.
+* Extended CarlaDataProvider's spawning functions to allow filtering the safer blueprint, and optionally tick the server
 
 ## CARLA ScenarioRunner 0.9.11
 ### :rocket: New Features
--- a/srunner/scenariomanager/carla_data_provider.py
+++ b/srunner/scenariomanager/carla_data_provider.py
@@ -620,6 +620,8 @@ class CarlaDataProvider(object):  # pylint: disable=too-many-public-methods
 
         actors = CarlaDataProvider.handle_actor_batch(batch, tick)
         for actor in actors:
+            if actor is None:
+                continue
             CarlaDataProvider._carla_actor_pool[actor.id] = actor
             CarlaDataProvider.register_actor(actor)
 
@@ -672,6 +674,8 @@ class CarlaDataProvider(object):  # pylint: disable=too-many-public-methods
 
         actors = CarlaDataProvider.handle_actor_batch(batch, tick)
         for actor in actors:
+            if actor is None:
+                continue
             CarlaDataProvider._carla_actor_pool[actor.id] = actor
             CarlaDataProvider.register_actor(actor)
 
--- a/srunner/scenariomanager/carla_data_provider.py
+++ b/srunner/scenariomanager/carla_data_provider.py
@@ -410,7 +410,7 @@ class CarlaDataProvider(object):  # pylint: disable=too-many-public-methods
         CarlaDataProvider._spawn_index = 0
 
     @staticmethod
-    def create_blueprint(model, rolename='scenario', color=None, actor_category="car"):
+    def create_blueprint(model, rolename='scenario', color=None, actor_category="car", safe=False):
         """
         Function to setup the blueprint of an actor given its model and other relevant parameters
         """
@@ -431,7 +431,22 @@ class CarlaDataProvider(object):  # pylint: disable=too-many-public-methods
 
         # Set the model
         try:
-            blueprint = CarlaDataProvider._rng.choice(CarlaDataProvider._blueprint_library.filter(model))
+            blueprints = CarlaDataProvider._blueprint_library.filter(model)
+            blueprints_ = []
+            if safe:
+                for bp in blueprints:
+                    if bp.id.endswith('isetta') or bp.id.endswith('carlacola') or bp.id.endswith('cybertruck') \
+                            or bp.id.endswith('t2') or bp.id.endswith('firetruck') or bp.id.endswith('ambulance'):
+                        # Remove troublesome blueprints (bulky / unstable)
+                        continue
+                    if int(bp.get_attribute('number_of_wheels')) != 4:
+                        # Two wheeled vehicles take much longer to render + bicicles shouldn't behave like cars
+                        continue
+                    blueprints_.append(bp)
+            else:
+                blueprints_ = blueprints
+
+            blueprint = CarlaDataProvider._rng.choice(blueprints_)
         except ValueError:
             # The model is not part of the blueprint library. Let's take a default one for the given category
             bp_filter = "vehicle.*"
@@ -508,11 +523,11 @@ class CarlaDataProvider(object):  # pylint: disable=too-many-public-methods
 
     @staticmethod
     def request_new_actor(model, spawn_point, rolename='scenario', autopilot=False,
-                          random_location=False, color=None, actor_category="car"):
+                          random_location=False, color=None, actor_category="car", safe=False):
         """
         This method tries to create a new actor, returning it if successful (None otherwise).
         """
-        blueprint = CarlaDataProvider.create_blueprint(model, rolename, color, actor_category)
+        blueprint = CarlaDataProvider.create_blueprint(model, rolename, color, actor_category, safe)
 
         if random_location:
             actor = None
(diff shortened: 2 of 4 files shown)


fabianoboril (changes requested):
> Reviewed 3 of 3 files at r1, 2 of 2 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/760)* status: all files reviewed, 5 unresolved discussions (waiting on @glopezdiest)

___
*[srunner/scenariomanager/carla_data_provider.py, line 442 at r3](https://reviewable.io/reviews/carla-simulator/scenario_runner/760#-MZsDl1e0QUHFuhjxqJO:-MZsDl1f3Kmwbz_Jo2dJ:b-rknn8r) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/7d841d0caaa8ccaf998777238f431a2c47ae0034/srunner/scenariomanager/carla_data_provider.py#L442)):*
> ```Python
>                         # Two wheeled vehicles take much longer to render + bicicles shouldn't behave like cars
>                         continue
>                     blueprints_.append(bp)
> ```

Hm... not sure if I should like that part? The leaderboard/CARLA doesn't like bikes so we remove bikes instead of fixing the root cause? Anyway I guess I have to accept it.
___
*[srunner/scenariomanager/carla_data_provider.py, line 543 at r3](https://reviewable.io/reviews/carla-simulator/scenario_runner/760#-MZsE4RY0JxrrTXeutOm:-MZsE4RY0JxrrTXeutOn:b-1c2sv3) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/7d841d0caaa8ccaf998777238f431a2c47ae0034/srunner/scenariomanager/carla_data_provider.py#L543)):*
> ```Python
>         if autopilot:
>             if actor.type_id.startswith('vehicle.'):
>                 actor.set_autopilot(autopilot, CarlaDataProvider._traffic_manager_port)
> ```

Is this a breaking API change? If so, we have to discuss when to merge this PR.
___
*[srunner/scenariomanager/carla_data_provider.py, line 625 at r3](https://reviewable.io/reviews/carla-simulator/scenario_runner/760#-MZsEBsQ9AEwzwbHmVGg:-MZsEBsQ9AEwzwbHmVGh:b8s16rg) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/9b244deab720e6e2e150f1f5b300cfecf352ba0b/srunner/scenariomanager/carla_data_provider.py#L625)):*
> ```Python
>         for actor in actors:
>             if actor is None:
>                 continue
> ```

I recall that there was a reason for this continue-statement. Maybe it is better to keep it at the moment.
___
*[srunner/scenariomanager/carla_data_provider.py, line 681 at r3](https://reviewable.io/reviews/carla-simulator/scenario_runner/760#-MZsEHecCgsBEwJlrkgR:-MZsEHecCgsBEwJlrkgS:b-tabvu0) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/9b244deab720e6e2e150f1f5b300cfecf352ba0b/srunner/scenariomanager/carla_data_provider.py#L681)):*
> ```Python
>         for actor in actors:
>             if actor is None:
>                 continue
> ```

See my previous comment
___
*[srunner/scenariomanager/result_writer.py, line 225 at r3](https://reviewable.io/reviews/carla-simulator/scenario_runner/760#-MZsEJQe1OpAOC1n5a8r:-MZsEJQe1OpAOC1n5a8s:b-ngldpw) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/7d841d0caaa8ccaf998777238f431a2c47ae0034/srunner/scenariomanager/result_writer.py#L225)):*
> ```Python
>             failure_count += 1
> 
>         with open(self._junit, "w") as junit_file:
> ```

What is this change good for?

glopezdiest (commented):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/760)* status: all files reviewed, 5 unresolved discussions (waiting on @fabianoboril)

___
*[srunner/scenariomanager/carla_data_provider.py, line 442 at r3](https://reviewable.io/reviews/carla-simulator/scenario_runner/760#-MZsDl1e0QUHFuhjxqJO:-MZvMcyz4mXbAi00aqLM:bauco65) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/7d841d0caaa8ccaf998777238f431a2c47ae0034/srunner/scenariomanager/carla_data_provider.py#L442)):*

Hm... not sure if I should like that part? The leaderboard/CARLA doesn't like bikes so we remove bikes instead of fixing the root cause? Anyway I guess I have to accept it.
</blockquote></details>

The fix would be to completely remake the bike blueprints. I truly believe that we should have a way to filter out bikes and motorbikes, as they behave different than other vehicles. The problem is that as the `create_blueprint` is currently being called from inside the spawning functions, it is very troublesome to  filter out specific blueprints.
___
*[srunner/scenariomanager/carla_data_provider.py, line 543 at r3](https://reviewable.io/reviews/carla-simulator/scenario_runner/760#-MZsE4RY0JxrrTXeutOm:-MZvMqSC7kH3Z5rAzrJ2:b-2erx9s) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/7d841d0caaa8ccaf998777238f431a2c47ae0034/srunner/scenariomanager/carla_data_provider.py#L543)):*

Is this a breaking API change? If so, we have to discuss when to merge this PR.
</blockquote></details>

No, the previous if statement was always false, as it was comparing a bool with a carla.Actor. Now that's not the case. I haven't seen any use of `request_new_actor` that actually tries to activate the autopilot
___
*[srunner/scenariomanager/carla_data_provider.py, line 625 at r3](https://reviewable.io/reviews/carla-simulator/scenario_runner/760#-MZsEBsQ9AEwzwbHmVGg:-MZvNltHEX9jUM528dkz:bbjmblr) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/9b244deab720e6e2e150f1f5b300cfecf352ba0b/srunner/scenariomanager/carla_data_provider.py#L625)):*

I recall that there was a reason for this continue-statement. Maybe it is better to keep it at the moment.
</blockquote></details>

As far as I can see, the handle_batch_actors will never return None, but I've readded it
___
*[srunner/scenariomanager/carla_data_provider.py, line 681 at r3](https://reviewable.io/reviews/carla-simulator/scenario_runner/760#-MZsEHecCgsBEwJlrkgR:-MZvNw1c97Xe7jxJbhjU:b-ya81ng) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/9b244deab720e6e2e150f1f5b300cfecf352ba0b/srunner/scenariomanager/carla_data_provider.py#L681)):*

See my previous comment
</blockquote></details>

Same as above
___
*[srunner/scenariomanager/result_writer.py, line 225 at r3](https://reviewable.io/reviews/carla-simulator/scenario_runner/760#-MZsEJQe1OpAOC1n5a8r:-MZvNxXTAo0yFX-vhklu:b-23eorb) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/7d841d0caaa8ccaf998777238f431a2c47ae0034/srunner/scenariomanager/result_writer.py#L225)):*

What is this change good for?
</blockquote></details>

None, but github was complaining about it, so I had to change it

fabianoboril (approved):
> Reviewed 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/760)* status: all files reviewed, 1 unresolved discussion (waiting on @glopezdiest)

___
*[srunner/scenariomanager/carla_data_provider.py, line 442 at r3](https://reviewable.io/reviews/carla-simulator/scenario_runner/760#-MZsDl1e0QUHFuhjxqJO:-MZvhHXt6yRw2UQrcKDj:b-vsvdpl) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/7d841d0caaa8ccaf998777238f431a2c47ae0034/srunner/scenariomanager/carla_data_provider.py#L442)):*
<details><summary><i>Previously, glopezdiest wrote…</i></summary><blockquote>

The fix would be to completely remake the bike blueprints. I truly believe that we should have a way to filter out bikes and motorbikes, as they behave different than other vehicles. The problem is that as the `create_blueprint` is currently being called from inside the spawning functions, it is very troublesome to  filter out specific blueprints.
</blockquote></details>

Agree. As said, I will accept this change, although I think that this should be handled in a different way.

glopezdiest:
Okay, we'll take about it, see if we can add more distinct tags to the blueprints (like *car*, *van*, bicicle*, etc. similar to our actor category). Also, can you help me with the static code? I can't seem to see what's failling

fabianoboril (approved):
> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/760)* status: all files reviewed, 1 unresolved discussion (waiting on @glopezdiest)

fabianoboril (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/760)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @glopezdiest)

PR #771, opened 2021-06-21
Fixed OSC LaneOffset sign handling

### LINKED ISSUE #767: Relative Lane Change directions are mixed up for OpenSCENARIO/OpenDRIVE
Hi,

I had issues running lane changes in OSC 1.0 Scenarios with OpenDRIVE maps. It turned out, the direction is interpreted differently compared to other tools like esmini.

In OpenDrive, the Lanes are defined in this way:
| 2 | 1 | -1 | -2 |

https://releases.asam.net/OpenDRIVE/1.6.0/ASAM_OpenDRIVE_BS_V1-6-0.html#_lanes

But in OpenSCENARIO parser, a negative relative lane target is interpreted as "left". But it should be "right" (as you can see in the above ODR definition).

https://github.com/carla-simulator/scenario_runner/blob/ad71a2c7ed012d735be2b1158fca51b0761ff26b/srunner/tools/openscenario_parser.py#L1032

I don't know, if a change here would bring implications for use-cases that do not load ODR maps.

Best regards
Simon

fabianoboril:
#Fixes #767

@sagar-g-v : I think changing the sign only for t < 0 is not correct, and has to be inverted always when using an common XODR coordinate system. Do you agree?

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/771)

### DIFF
--- a/Docs/CHANGELOG.md
+++ b/Docs/CHANGELOG.md
@@ -23,6 +23,7 @@
 * Additional Scenarios:
     - Added Construction setup scenario.
 ### :bug: Bug Fixes
+* Fixed LaneOffset (+ vs. -) for OpenSCENARIO
 * Fixed bug at the Getting Started docs which caused an import error
 * Fixed neverending lane change maneuver in OpenSCENARIO
 * Fixed bug causing the spawning of an actor with `request_new_actor` to never activate the autopilot.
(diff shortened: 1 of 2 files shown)


sagar-g-v:
> #Fixes #767
> 
> @sagar-g-v : I think changing the sign only for t < 0 is not correct, and has to be inverted always when using an common XODR coordinate system. Do you agree?
> 
> This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/771)

Yes. I agree with this

fpasch (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/771)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @fabianoboril)

PR #837, opened 2021-11-12
Improved scenarios

glopezdiest:
### Description

Many of the python scenarios have been improved, as they were quite outdated. The focus was mostly on the scenarios part of the routes, making them more robust, more general to any configuration and interesting. The changes scenarios are the following:

- **ControlLoss** (control_loss.py): The code was pretty outdated, duplicated and used behaviors such as the control noise, which didn't do anything. That has been fixed, and debris are now projected onto the ground, avoiding some floating debris in some parts of Town03.
- **FollowLeadingVehicle** (follow_leading_vehicle.py): The other vehicle is no longer moved below ground, and moved back when the ego moves. Added a new Route version that utilizes the new background activity
- **OtherLeadingVehicle** (other_leading_vehicle.py): There was a missing conversion from Km/h to m/s, causing the other actors to move way too fast
- **VehicleTurning** (object_crash_intersection.py): Created a new BaseVehicleTurning (as all code but one line was the exact same. The subscenarios (Left, Right and Route) depend on this class. The crossing actors no longer move below ground and respawn when the ego is closeby. Updated synchronization and overall code (which was a bit outdated) 
- **DynamicObjectCrossing** (object_crash_vehicle.py): Updated synchronization and overall code (which was a bit outdated). The other actors (pedestrian and blocker) no longer move below ground and appear once the ego is closeby 
- **RunningRedLight** (opposite_vehicle_taking_priority.py): Complete remake. The collision point is calculated and used to synchronize the other actor, which will traverse the intersection at a constant velocity. It now only needs the position of the ego vehicle and will randomly choose between ambulance / firetruck / police, which will come from either the left or the right. Now freezes the traffic lights of the intersection
- **SignalizedJunctionLeftTurn** (signalized_junction_left_turn.py): Generalized for better usability at routes. Improved actor flow. Now freezes the traffic lights of the intersection
- **SignalizedJunctionRightTurn** (signalized_junction_right_turn.py): Remade. Instead of a synchronization, it is now has an actor flow, going from left to right (similar to SignalizedJucntionLeftTurn). The ego is meant to smoothly enter the flow. Now freezes the traffic lights of the intersection
- **BackgroundActivity** (background_activity.py): Complete remake. It is now an scenario, not just some random vehicles controlled by the TM, meant to create the sensation of traffic around the ego. Added a background_manager.py that allows the user to modify its parameters on real time. This is meant as a way to stop the background activity from interfering with other scenarios.

In order to create the sensation of traffic around the ego, the new **BackgroundActivity** handles the creation and destruction of vehicles around it. This allows for more interesting driving scenarios, a diverse environment around the ego and an overall performance increase, as vehicles are only spawned around the ego, and not anywhere else. The vehicles are still controlled by the TM.

The BackgroundActivity has **two modes**, road mode for when the ego has no junctions nearby and junction mode, which starts when there is an junction closeby and ends when the ego exits that junction. For two or more junctions next to each other, they can all be active at the same time, and the road mode will only activate once all junctions have been finished. Only junctions part of the route will be activated.

**Road mode** spawns a certain amount of vehicles in front and behind the ego (for all lanes in the same direction). The vehicles in front will reduce speed when far away, maintaining the flow of traffic. Additionally, the opposite lanes also have traffic. When a junction starts, these actors will be automatically part of the junction mode

**Junction mode** creates actor sources at all entries that are not part of the route, and actor sinks at all exits. Actor sinks don't directly remove vehicles but instead track the amount of actors at a specific lane, destroying the farthest one when a critical number is reached. Similar to the road behavior, actors will reduce speed if they are too far from the jucntion. To ensure the minimum amount of road vehicles is always available, the road through which the route exits is populated from the start. When the ego exits the junction, all actors part of the junction mode will be destroyed, except those part of the route's road, which will serve as the new road mode actors. Source will be added at the end of the junction, which will eventually create the vehicels behind the ego. For multiple junctions, actors that exit one will automatically be part of the entry actors of the next.

In case this all felt like gibberish, **here is a video** of the new BackgroundActivity. At the first junction, normal BA behavior is seen. The second and third junctions are activated at the same time, and the actor destruction is clearly shown when exiting the second one. The third one shows how the SignalizedJucntionTurningLeft scenario is triggered and the background activity changes to no interfere with the actor flow (removing the actor source and exit on which the flow moves and destroying all BA actors that enter the junction)

https://user-images.githubusercontent.com/58212725/141466721-04b3937a-8c03-4903-aed5-eb70a10e98e4.mp4

Moreover, the ***.json* scenario file have been cleaned up and improved**. *all_towns_traffic_scenarios1_3_4_8.json* and *all_towns_traffic_scenarios1_3_4.json* have been removed, as they were subsets of *all_towns_traffic_scenarios.json*. This last file has had many duplicate scenarios removed, others have had their position slightly modified (as they were misplaced or too high to be triggered), and scenarios 2, 7, 8, 9 and 10 have been added to all towns.

**Other minor changes**:
- Removed some hardcoded values at **human_agent.py** and improved the code at **npc_agent.py**.
- The **CarlaDataProvider seed** has been exposed. This is so that at routes, scenarios can be random and deterministic at the same time, as the other option, using a seed, made the scenario repeat itself throughout all the route
- **Remade how the route is calculated**. As it was done in chunks, lane changes between the end of a chunk and the start of the next had the incorrect RoadOption. it now creates the full route (which is the same as before) and then adds the RoadOptions.

### Where has this been tested?

  * **Python version(s):** 3.6
  * **CARLA version:** 0.9.12


fabianoboril (commented):
> Reviewed 28 of 28 files at r1.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/837)* status: all files reviewed (commit messages unreviewed), 4 unresolved discussions (waiting on @glopezdiest)

___
*[srunner/scenariomanager/scenarioatomics/atomic_behaviors.py, line 1544 at r1](https://reviewable.io/reviews/carla-simulator/scenario_runner/837#-MoJIuILDYh3M4DnHt4E:-MoJIuILDYh3M4DnHt4F:be5o1o7) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/3516a0f8d6f535ee281b9cef53f9efb27005156d/srunner/scenariomanager/scenarioatomics/atomic_behaviors.py#L1544)):*
> ```
>             self._control.manual_gear_shift = True
>             self._control.gear = 4
> ```
Hm... This is some really dirty "hack".. Why not using something similar to what we use for the initial speed setting, i.e. use set_target_velocity instead of using a real controller?
___
*[srunner/scenariomanager/scenarioatomics/atomic_behaviors.py, line 2720 at r1](https://reviewable.io/reviews/carla-simulator/scenario_runner/837#-MoJJYpV8BL1py3SbRDx:-MoJJYpV8BL1py3SbRDy:b-3e7zec) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/3516a0f8d6f535ee281b9cef53f9efb27005156d/srunner/scenariomanager/scenarioatomics/atomic_behaviors.py#L2720)):*
> ```Python
>         """
>         Sets the traffic lights to the desired states and remembers their previous one.
>         These should technically be frozen, but that freezes all tls in the town
> ```

Hm... This sounds to me like the requirement for a new CARLA command ;)
___
*[srunner/scenarios/basic_scenario.py, line 44 at r1](https://reviewable.io/reviews/carla-simulator/scenario_runner/837#-MoJJjsQ19S8aTk-SIAc:-MoJJjsQ19S8aTk-SIAd:br14x2j) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/3516a0f8d6f535ee281b9cef53f9efb27005156d/srunner/scenarios/basic_scenario.py#L44)):*
> ```Python
>         self.criteria_list = []  # List of evaluation criteria
>         self.scenario = None
>         self.world = world
> ```

The world can be fetched via the carla_data_provider. Why is it needed to become part of the scenario?
___
*[srunner/tools/scenario_helper.py, line 580 at r1](https://reviewable.io/reviews/carla-simulator/scenario_runner/837#-MoJKRQ8EfjXisYI0s2H:-MoJKRQ8EfjXisYI0s2I:b2c8po5) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/3516a0f8d6f535ee281b9cef53f9efb27005156d/srunner/tools/scenario_helper.py#L580)):*
> ```Python
> 
> 
> def get_closest_traffic_light(waypoint, traffic_lights=None):
> ```

Do you want to get the closest, or the closest relevant one? This is not necessarily the same. In general, this might be better handled in CARLA, with a new API function.

glopezdiest (commented):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/837)* status: 15 of 29 files reviewed, 4 unresolved discussions (waiting on @fabianoboril)

___
*[srunner/scenariomanager/scenarioatomics/atomic_behaviors.py, line 1544 at r1](https://reviewable.io/reviews/carla-simulator/scenario_runner/837#-MoJIuILDYh3M4DnHt4E:-MoY0XeJ3a3GkT6W9K_b:b-cw1o1u) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/3516a0f8d6f535ee281b9cef53f9efb27005156d/srunner/scenariomanager/scenarioatomics/atomic_behaviors.py#L1544)):*

> ```
>             self._control.manual_gear_shift = True
>             self._control.gear = 4
> ```
Hm... This is some really dirty "hack".. Why not using something similar to what we use for the initial speed setting, i.e. use set_target_velocity instead of using a real controller?
</blockquote></details>

Yeah, it was an old behavior. The plan was to maintain constant speed at high speed but  I just couldn't use the init target velocity as the engine was stopped, causing it to try to be at high speed with the first gear. I figured out another way to do so, so I'm removing it
___
*[srunner/scenariomanager/scenarioatomics/atomic_behaviors.py, line 2720 at r1](https://reviewable.io/reviews/carla-simulator/scenario_runner/837#-MoJJYpV8BL1py3SbRDx:-MoXbfMR7uZUHFYHPXUo:bhqunj2) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/3516a0f8d6f535ee281b9cef53f9efb27005156d/srunner/scenariomanager/scenarioatomics/atomic_behaviors.py#L2720)):*

Hm... This sounds to me like the requirement for a new CARLA command ;)
</blockquote></details>

I agree, but it was decided against it a while back
___
*[srunner/scenarios/basic_scenario.py, line 44 at r1](https://reviewable.io/reviews/carla-simulator/scenario_runner/837#-MoJJjsQ19S8aTk-SIAc:-MoXbqoeBUXej6R2pq1m:bz3fb2e) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/3516a0f8d6f535ee281b9cef53f9efb27005156d/srunner/scenarios/basic_scenario.py#L44)):*

The world can be fetched via the carla_data_provider. Why is it needed to become part of the scenario?
</blockquote></details>

Sure you can, but we are already passing it to the scenario as an argument, so it feels more natural to me to just use that one.
___
*[srunner/tools/scenario_helper.py, line 580 at r1](https://reviewable.io/reviews/carla-simulator/scenario_runner/837#-MoJKRQ8EfjXisYI0s2H:-MoXemxn4r4lPlZH76gv:b-d1gozm) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/3516a0f8d6f535ee281b9cef53f9efb27005156d/srunner/tools/scenario_helper.py#L580)):*
Technically the closest relevant one. The idea was, given a waypoint in the entrance of the junction, get the traffic light affecting that lane. However, checking that relevance is not that straightforward, and might cause false negatives.

> In general, this might be better handled in CARLA, with a new API function.

Yeah, but right now its not :) We'll see if it is worth upgrading the waypoint API

joel-mb (approved):
> Reviewed 13 of 28 files at r1, 12 of 15 files at r2, 1 of 2 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/837)* status: all files reviewed, 4 unresolved discussions (waiting on @fabianoboril)

fabianoboril (approved):
> Reviewed 12 of 15 files at r2, 1 of 2 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/837)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @glopezdiest)

PR #866, opened 2022-03-08
Parsing times to compute speed

jackokaiser:
#### Description

Reading times on `FollowTrajectoryAction` Vertex elements to set the speed along the trajectory. Vertex times are assumed relative and TimeReference is ignored. Speed is linearly interpolated between waypoints. 

[Here](https://github.com/carla-simulator/scenario_runner/files/8206723/FollowTrajectoryTest.zip) is the updated example to test the FollowTrajectoryAction.

  * **Platform(s):** linux
  * **Python version(s):** 3.8.12
  * **Unreal Engine version(s):**  4.26.2-0+++UE4+Release-4.26 522 0
  * **CARLA version:** 0.9.13

#### Possible Drawbacks

- added a positional argument to ChangeActorWaypoints's init function 
- vehicles are not teleported when they should, according to OpenScenario documentation (see 3.7.2. in OpenScenario v1.1.1)

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/866)

### DIFF
--- a/srunner/scenariomanager/actorcontrols/actor_control.py
+++ b/srunner/scenariomanager/actorcontrols/actor_control.py
@@ -158,15 +158,6 @@ class ActorControl(object):
         """
         return self._last_waypoint_command
 
-    def get_waypoints(self):
-        """
-        Get the remaining waypoints
-
-        returns:
-            The remaining waypoints
-        """
-        return self.control_instance.get_waypoints()
-
     def get_last_lane_offset_command(self):
         """
         Get timestamp of the last issued lane offset control command
--- a/srunner/scenariomanager/actorcontrols/actor_control.py
+++ b/srunner/scenariomanager/actorcontrols/actor_control.py
@@ -158,6 +158,15 @@ class ActorControl(object):
         """
         return self._last_waypoint_command
 
+    def get_waypoints(self):
+        """
+        Get the remaining waypoints
+
+        returns:
+            The remaining waypoints
+        """
+        return self.control_instance.get_waypoints()
+
     def get_last_lane_offset_command(self):
         """
         Get timestamp of the last issued lane offset control command
--- a/srunner/scenariomanager/actorcontrols/npc_vehicle_control.py
+++ b/srunner/scenariomanager/actorcontrols/npc_vehicle_control.py
@@ -137,6 +137,3 @@ class NpcVehicleControl(BasicControl):
             light_state = self._actor.get_light_state()
             light_state &= ~carla.VehicleLightState.Brake
             self._actor.set_light_state(carla.VehicleLightState(light_state))
-
-    def get_waypoints(self):
-        return self._local_planner._waypoints_queue
--- a/srunner/scenariomanager/actorcontrols/npc_vehicle_control.py
+++ b/srunner/scenariomanager/actorcontrols/npc_vehicle_control.py
@@ -93,7 +93,6 @@ class NpcVehicleControl(BasicControl):
 
         if self._waypoints_updated:
             self._waypoints_updated = False
-            self._local_planner.num_reached_waypoints = 0
             self._update_plan()
 
         if self._offset_updated:
--- a/srunner/scenariomanager/actorcontrols/npc_vehicle_control.py
+++ b/srunner/scenariomanager/actorcontrols/npc_vehicle_control.py
@@ -93,6 +93,7 @@ class NpcVehicleControl(BasicControl):
 
         if self._waypoints_updated:
             self._waypoints_updated = False
+            self._local_planner.num_reached_waypoints = 0
             self._update_plan()
 
         if self._offset_updated:
@@ -137,3 +138,6 @@ class NpcVehicleControl(BasicControl):
             light_state = self._actor.get_light_state()
             light_state &= ~carla.VehicleLightState.Brake
             self._actor.set_light_state(carla.VehicleLightState(light_state))
+
+    def get_waypoints(self):
+        return self._local_planner._waypoints_queue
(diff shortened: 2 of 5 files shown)


glopezdiest:
Hey @jackokaiser. this looks good to me. Simple yet effective.

My only concern is with how you calculate the speed on runtime. Your initial computation and interpolation creates a locked trajectory the vehicle is meant to follow but what if, for some reason (controller, collisions / physics ...), it doesn't? I think, an initial interpolation isn't enough, and we need some kind of closed loop on runtime that gets the distance to the next waypoint, and from there, calculates the speed.

jackokaiser:
Hi @glopezdiest,

Thanks for your quick answer!

It's a good idea to calculate the speed in closed-loop, with respect to the current location and the next waypoint to reach. The implementation isn't so straightforward though, because it's hard to retrieve what waypoint we are trying to reach from the AtomicBehavior.

To resolve this problem I've added a new method `get_waypoints`, which should be implemented by the control instance of `ActorControl`. I provided the implementation for PedestrianControl and NpcVehicleControl. Here are the drawbacks of this approach:
- `get_waypoints` should be implemented by all Control class, willing to make use of FollowTrajectoryAction
- the method returns a "private" member variable by reference (the list/points could be modified externally)  

Note that it isn't enough to just return the next waypoint, as this would lead to an association problem: we need to know the current waypoint index, to know its time to arrival. Alternatively, we could have `get_num_waypoints` instead of  `get_waypoints`, as it would be sufficient to know how many waypoints remain to be reached. 

I've also looked into WaypointFollower which could replace ChangeActorWaypoints, because the latter has it own planner, but it requires to temporarily remove the actor from ActorsWithController in the Blackboard, and this isn't suited for pedestrians.

glopezdiest:
Yeah, it is true that it is quite hard to create a closed loop. Thanks for the update, I'll take a look ASAP, see how that works. 

> I've also looked into WaypointFollower which could replace ChangeActorWaypoints, because the latter has it own planner, but it requires to temporarily remove the actor from ActorsWithController in the Blackboard, and this isn't suited for pedestrians.

Don't look into WaypointFollowe, that's the old version we were using and it has lots of issues for SR. These actor controls are the solution that we ended up creating.

glopezdiest (commented):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/866)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @glopezdiest and @jackokaiser)

___
*[srunner/scenariomanager/scenarioatomics/atomic_behaviors.py, line 879 at r3](https://reviewable.io/reviews/carla-simulator/scenario_runner/866#-MyvfDje3pVMZyNLGWkO:-MyvfDjf47ovSa-6IK8f:b4s3g89) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/89c8e76fb8c93be0bf18a72b8a50e23fd671199b/srunner/scenariomanager/scenarioatomics/atomic_behaviors.py#L879)):*
> ```Python
>         if 0 <= waypoint_idx < len(self._times):
>             target_waypoint = remaining_waypoints[0]
>             if isinstance(target_waypoint, carla.Transform):
> ```

This is not working, as the waypoints of the controllers aren't the same as the ones you input, so your times are incorrect.

You have  to compare your position to that of the waypoints given to the behavior (self._waypoints), and add a check to see if the actor has passed that waypoint. We do something similar at one of our criteria, using the dot product ([here](https://github.com/carla-simulator/scenario_runner/blob/master/srunner/scenariomanager/scenarioatomics/atomic_criteria.py#L1651))



_Code quote:_
```Python
            target_waypoint = remaining_waypoints[0]
            if isinstance(target_waypoint, carla.Transform):
                target_location = target_waypoint.location
            elif isinstance(target_waypoint, tuple) and isinstance(target_waypoint[0], carla.Waypoint):
                target_location = target_waypoint[0].transform.location
            else:
                raise RuntimeError(f'Unexpected waypoint type: {type(target_waypoint)}')
```

jackokaiser:
Hi @glopezdiest,

On a second look at the OpenSCENARIO documentation, I feel that the expected behavior for an agent not able to reach a waypoint on time would simply be to drive towards the next waypoints rather than sticking to an old waypoint it was unable to reach.

Quotes from the documentation (for "follow" mode):

> 3.7.2. Entity is not at the start of trajectory, timeReference is set (in the future)
> Expected behavior: the entity immediately starts steering towards the trajectory, with the aim to be at the trajectory start at t=t1.

> 3.7.3. Entity is not at the start of trajectory, timeReference is set (in the past)
> Expected behavior: the entity starts steering towards the trajectory, aiming to hit the first referenced point in the future and taking into consideration the performance limitations

If you agree with this interpretation, I will propose a new implementation for such behavior. For example, the waypoints could be fed one by one to the actor, based on the time reference, in the update function of `ChangeActorWaypoints`

glopezdiest:
Yeah, you are right. The way I see it, after reaching that time, waypoint are ignored, regardless of whether you managed to reach them or not.

glopezdiest:
Does the last commit implement these changes?

glopezdiest (approved):
> Reviewed 2 of 5 files at r2, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/866)* status: all files reviewed, 1 unresolved discussion (waiting on @jackokaiser)

glopezdiest (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/866)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jackokaiser)

glopezdiest:
Merged, this looks great.

jackokaiser:
nice, thanks

bapukokare (commented):
follow trajectory action happens in the same lane. if the trajectory contains with lane-change action information (change in x,y value), NPC control does not follow that.   how to increase the support to lane change action which have vertices with  x,y information, go straight action 

> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/866#-:-NIXzYGf01XYh6d_eS2D:b9oa5yk)* status: :shipit: complete! all files reviewed; all discussions resolved

PR #874, opened 2022-03-24
Added dynamic weathers to routes

glopezdiest:
#### Description

Added dynamic weather to routes. Weathers are now set at specific route positions, and the simulation will automatically switch between them. An example would be something like this:

image

All weathers need a `route_percentage`, which is the route position previously mentioned, and is expressed as a percentage of the route. The rest of the parameters are the *carla.WeatherParameters* attribute that we want to change (not all are needed). 
During the simulation, the position of the ego vehicle will be monitored, and everytime it advances through the route, the weather will be changed, interpolating between the two nearest weather keypoints.

For routes with one single weather, the weather will remain constant throughout the simulation, and if the extreme cases are not defined (0% and 100% positions), all positions inside the intervals (i.e from 0% until the defined position at 10%) will have the same weather.

Here is a video showing an example:

https://user-images.githubusercontent.com/58212725/159895922-53c5a00a-32de-415e-981e-a1be7e75206f.mp4

  * **Platform(s):** Ubutnu 18.04
  * **Python version(s):** 3.6
  * **Unreal Engine version(s):** CARLA's
  * **CARLA version:** 0.9.13

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/874)

### DIFF
--- a/Docs/CHANGELOG.md
+++ b/Docs/CHANGELOG.md
@@ -31,6 +31,7 @@
 * Scenarios are no logner fixed to a location but instead are now part of the route.
 * Routes now automatically import all scenarios.
 * Routes will now also take into account the criteria of their specific scenarios, only being active when they are running. The ResultWriter now automaically adds these criteria, grouping them if more than one scenario of the same type is triggered.
+* Routes can now have dynamic weather. These are set by keypoints at a routes percentage, and all values between them are interpolated.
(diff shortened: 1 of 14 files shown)


jackbart94 (approved):
> Reviewed 9 of 9 files at r1, 1 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/874)* status: 7 of 10 files reviewed, all discussions resolved (waiting on @jackbart94)

PR #878, opened 2022-04-12
Fast Exit scenario

TayYim:
#### Description

Add Fast Exit scenario. 

Based on CrossActorFlow, I add a parameter called "clear_lane". If it's true, then the scenario will clear the Background Activity on the lane where the actor flow starts.

An AtomicBehavior called RemoveLane is added to background_manager.py. The clear_lane action will be executed in _update_parameters method in background_activity.py.

A route file, FastExit.xml, is uploaded to examples.

Fixes #  

  * **Python version(s):** 3.7
  * **Unreal Engine version(s):4.26

#### Possible Drawbacks

Didn't test with other scenarios, don't know whether it is compatible with others.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/878)


glopezdiest:
I'd recommend adding the `FastExit.xml` into a new folder, as this is just a temporary route created to test the scenario. We can place all the testing rotues there, and once the Leaderboard 2.0 is finished, remove the folder. 

Additionally, don't overwrite the CrossActorFlow scenario. By doing so, this new fast exit scenario is forced to have the `clear_lane` argument, which in reality it is part of the scenario behavior, as without it, the scenario breaks. 

Some comments about the implementations:

- The way this lane removal works is by checking the `road_lane_key`, which is only available when the BA works in road mode (active when there are no junctions in front of the ego). Is this a known limitation?
- Your example shows the actor flow starting right at the side of the ego. This is due to `start_actor_flow` parameter and can be easily changed, but it is very ugly to allow the ego to see spawning vehicles. They should be spawned some distance being it

TayYim:
New updates:
1. Put route file in /srunner/routes folder.
2. The ActorFlow starts far behind the EGO. 
3. Create a new scenario in actor_flow.py called CleanCrossActorFlow. It's almost the same as CrossActorFlow.
4. It is. It works when BA is working and there're data in _road_dict object.
> The way this lane removal works is by checking the `road_lane_key`, which is only available when the BA works in road mode (active when there are no junctions in front of the ego). Is this a known limitation?

srunner/scenarios/actor_flow.py
```diff
@@ -319,3 +328,93 @@ def __del__(self):
...
+            source_wp, sink_wp, self._source_dist_interval, self._sink_distance, self._flow_speed))
+        end_condition = py_trees.composites.Sequence()
+        end_condition.add_child(WaitEndIntersection(self.ego_vehicles[0], junction_id))
+        
+        root.add_child(end_condition)
+
+        sequence = py_trees.composites.Sequence()
+
+        if self.route_mode:
+            sequence.add_child(RemoveLane(str(source_wp.lane_id)))
```
glopezdiest, line 402: Why is the id a string?

srunner/scenarios/actor_flow.py
```diff
@@ -319,3 +328,93 @@ def __del__(self):
...
+            sequence.add_child(JunctionScenarioManager('opposite', True))
+            sequence.add_child(StopEntries())
```
glopezdiest, line 404: Up to discussion, but the StopEntries might not be needed for this type of scenario, as it stops all non route entries, which the highway doesn't have, as the intersection is just the highway exit.

srunner/scenarios/actor_flow.py
```diff
@@ -319,3 +328,93 @@ def __del__(self):
...
+        A list of all test criteria will be created that is later used
+        in parallel behavior tree.
+        """
+        return [CollisionTest(self.ego_vehicles[0])]
+
+    def __del__(self):
+        """
+        Remove all actors and traffic lights upon deletion
+        """
+        self.remove_all_actors()
```
glopezdiest, line 420: Add an empty line at the end of the file

srunner/tools/background_manager.py
```diff
@@ -235,3 +235,20 @@ def update(self):
...
+        lane_id (str): A carla.Waypoint.lane_id
+    """
+    def __init__(self, lane=None, name="RemoveLane"):
+        self._lane = lane
+        super().__init__(name)
+
+    def update(self):
+        """Updates the blackboard and succeds"""
+        py_trees.blackboard.Blackboard().set("BA_RemoveLane", self._lane, overwrite=True)
+        return py_trees.common.Status.SUCCESS
```
glopezdiest, line 254: End file with empty line

srunner/scenarios/actor_flow.py
```diff
@@ -319,3 +328,93 @@ def __del__(self):
         Remove all actors and traffic lights upon deletion
         """
         self.remove_all_actors()
+
+class CleanCrossActorFlow(BasicScenario):
```
glopezdiest, line 332: This needs a more clear name to understand the expected usecase of the scenario. Something like FastHighwayExit? Does this work if the exit is on the right? If so, then just HighwayExit?

srunner/scenarios/actor_flow.py
```diff
@@ -319,3 +328,93 @@ def __del__(self):
...
+    """
+    This scenario is similar to CrossActorFlow
```
glopezdiest, line 334: Similar to the name, the docs should reflect the expected behavior of the scenario, along with its limitations

glopezdiest (changes requested):

srunner/scenarios/actor_flow.py
TayYim: It works if the exit is on the right. HighwayExit is fine.

srunner/scenarios/actor_flow.py
TayYim: I didn't notice this. I think both JunctionScenarioManager and StopEntries are unnecessary since this scenario has no junctions.

srunner/scenarios/actor_flow.py
TayYim: It doesn't need to be cast into string here. Removed.

srunner/scenarios/background_activity.py
```diff
@@ -1631,6 +1631,19 @@ def _update_parameters(self):
             self._leave_space_in_front(leave_space_data)
             py_trees.blackboard.Blackboard().set('BA_LeaveSpaceInFront', None, True)
 
+        # Remove Lane
+        remove_lane_data = py_trees.blackboard.Blackboard().get('BA_RemoveLane')
+        if remove_lane_data is not None and isinstance(remove_lane_data, str):
+            lane_id_list = [x.split('*')[-1] for x in  list(self._road_dict.keys())]
```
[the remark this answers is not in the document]
TayYim: Put this in _remove_lane(self, lane_id) method in BackgroundBehavior Class

glopezdiest (approved):
Looks good!

glopezdiest (approved):
> Reviewed 1 of 1 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/878)* status: all files reviewed, 9 unresolved discussions (waiting on @glopezdiest and @TayYim)

glopezdiest (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/878)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @TayYim)

PR #882, opened 2022-04-25
add Parking Exit scenario

TayYim:
#### Description

Add Parking Exit scenario. The scenario in which the ego is parked between two vehicles and has to maneuver to start the route.

  * **Python version(s):** 3.7

#### Possible Drawbacks

1. Need to discuss: the distance leaving for blocking vehicles in front of and behind the parking point. Currently it is configurable

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/882)

### DIFF
--- a/srunner/scenariomanager/scenarioatomics/atomic_criteria.py
+++ b/srunner/scenariomanager/scenarioatomics/atomic_criteria.py
@@ -1020,19 +1020,6 @@ class OutsideRouteLanesTest(Criterion):
         if location is None:
             return new_status
 
-        # Deactivate by other components
-        activate = py_trees.blackboard.Blackboard().get('AC_SwitchOutsideRouteLanesTest')
-        if activate is not None and activate is False:
-            # Don't check if it's deactivated
-            pass
-        else:
-            # Check if it's activated
-            self._is_outside_driving_lanes(location)
-            self._is_at_wrong_lane(location)
-
-        if self._outside_lane_active or self._wrong_lane_active:
-            self.test_status = "FAILURE"
-
         # Get the traveled distance
         for index in range(self._current_index + 1,
                            min(self._current_index + self.WINDOWS_SIZE + 1, self._route_length)):
@@ -1054,6 +1041,20 @@ class OutsideRouteLanesTest(Criterion):
                     self._wrong_distance += new_dist
 
                 self._current_index = index
+
+        # Deactivate/Activate cheching by blackboard message
+        activate = py_trees.blackboard.Blackboard().get('AC_SwitchOutsideRouteLanesTest')
+        if activate is not None and activate is False:
+            # Don't check if it's deactivated
+            pass
+        else:
+            # Check if it's activated
+            self._is_outside_driving_lanes(location)
+            self._is_at_wrong_lane(location)
+        py_trees.blackboard.Blackboard().set('AC_SwitchOutsideRouteLanesTest', None, True)
+
+        if self._outside_lane_active or self._wrong_lane_active:
+            self.test_status = "FAILURE"
 
         self.logger.debug("%s.update()[%s->%s]" % (self.__class__.__name__, self.status, new_status))
         return new_status
--- a/srunner/scenariomanager/scenarioatomics/atomic_criteria.py
+++ b/srunner/scenariomanager/scenarioatomics/atomic_criteria.py
@@ -1020,9 +1020,15 @@ class OutsideRouteLanesTest(Criterion):
         if location is None:
             return new_status
 
-        # Check if outside route lanes
-        self._is_outside_driving_lanes(location)
-        self._is_at_wrong_lane(location)
+        # Deactivate by other components
+        activate = py_trees.blackboard.Blackboard().get('AC_SwitchOutsideRouteLanesTest')
+        if activate is not None and activate is False:
+            # Don't check if it's deactivated
+            pass
+        else:
+            # Check if it's activated
+            self._is_outside_driving_lanes(location)
+            self._is_at_wrong_lane(location)
 
         if self._outside_lane_active or self._wrong_lane_active:
             self.test_status = "FAILURE"
--- a/srunner/scenariomanager/scenarioatomics/atomic_criteria.py
+++ b/srunner/scenariomanager/scenarioatomics/atomic_criteria.py
@@ -1003,6 +1003,7 @@ class OutsideRouteLanesTest(Criterion):
         self._last_lane_id = None
         self._total_distance = 0
         self._wrong_distance = 0
+        self._active = True
 
     def update(self):
         """
@@ -1019,6 +1020,18 @@ class OutsideRouteLanesTest(Criterion):
         location = CarlaDataProvider.get_location(self.actor)
         if location is None:
             return new_status
+
+        # Deactivate/Activate checking by blackboard message
+        active = py_trees.blackboard.Blackboard().get('AC_SwitchOutsideRouteLanesTest')
+        if active is not None:
+            self._active = active
+            py_trees.blackboard.Blackboard().set("AC_SwitchOutsideRouteLanesTest", None, overwrite=True)
+
+        self._is_outside_driving_lanes(location)
+        self._is_at_wrong_lane(location)
+
+        if self._active and (self._outside_lane_active or self._wrong_lane_active):
+            self.test_status = "FAILURE"
 
         # Get the traveled distance
         for index in range(self._current_index + 1,
@@ -1037,24 +1050,11 @@ class OutsideRouteLanesTest(Criterion):
                 self._total_distance += new_dist
 
                 # And to the wrong one if outside route lanes
-                if self._outside_lane_active or self._wrong_lane_active:
+                if self._active and (self._outside_lane_active or self._wrong_lane_active):
                     self._wrong_distance += new_dist
 
                 self._current_index = index
 
-        # Deactivate/Activate cheching by blackboard message
-        activate = py_trees.blackboard.Blackboard().get('AC_SwitchOutsideRouteLanesTest')
-        if activate is not None and activate is False:
-            # Don't check if it's deactivated
-            pass
-        else:
-            # Check if it's activated
-            self._is_outside_driving_lanes(location)
-            self._is_at_wrong_lane(location)
-        py_trees.blackboard.Blackboard().set('AC_SwitchOutsideRouteLanesTest', None, True)
-
-        if self._outside_lane_active or self._wrong_lane_active:
-            self.test_status = "FAILURE"
 
         self.logger.debug("%s.update()[%s->%s]" % (self.__class__.__name__, self.status, new_status))
         return new_status
--- a/srunner/scenarios/parking_exit.py
+++ b/srunner/scenarios/parking_exit.py
@@ -98,7 +98,6 @@ class ParkingExit(BasicScenario):
         sequence = py_trees.composites.Sequence()
 
         # Put blocking vehicles
-        # parking_waypoint = self._parking_waypoint
         if self._parking_lane_side == "left":
             parking_waypoint = self._map.get_waypoint(
                 self.ego_vehicles[0].get_location()).get_left_lane()
--- a/srunner/scenarios/parking_exit.py
+++ b/srunner/scenarios/parking_exit.py
@@ -76,6 +76,18 @@ class ParkingExit(BasicScenario):
         else:
             self._parking_lane_side = "right"
 
+        # Get parking_waypoint based on trigger_point
+        self._trigger_location = config.trigger_points[0].location
+        self._reference_waypoint = self._map.get_waypoint(self._trigger_location)
+        if self._parking_lane_side == "left":
+            self._parking_waypoint = self._reference_waypoint.get_left_lane()
+        else:
+            self._parking_waypoint = self._reference_waypoint.get_right_lane()
+
+        if self._parking_waypoint is None:
+            raise Exception(
+                "Couldn't find parking point on the {} side".format(self._parking_lane_side))
+
         super(ParkingExit, self).__init__("ParkingExit",
                                           ego_vehicles,
                                           config,
@@ -88,17 +100,6 @@ class ParkingExit(BasicScenario):
         Custom initialization
         """
         # Put blocking vehicles
-        if self._parking_lane_side == "left":
-            self._parking_waypoint = self._map.get_waypoint(
-                self.ego_vehicles[0].get_location()).get_left_lane()
-        else:
-            self._parking_waypoint = self._map.get_waypoint(
-                self.ego_vehicles[0].get_location()).get_right_lane()
-
-        if self._parking_waypoint is None:
-            raise Exception(
-                "Couldn't find parking point on the {} side".format(self._parking_lane_side))
-
         front_points = self._parking_waypoint.next(
             self._front_vehicle_distance)
         if front_points:
--- a/srunner/scenarios/parking_exit.py
+++ b/srunner/scenarios/parking_exit.py
@@ -1,0 +1,174 @@
+#!/usr/bin/env python
+
+# Copyright (c) 2018-2020 Intel Corporation
+#
+# This work is licensed under the terms of the MIT license.
+# For a copy, see <https://opensource.org/licenses/MIT>.
+
+"""
+Scenario in which the ego is parked between two vehicles and has to maneuver to start the route.
+"""
+
+from __future__ import print_function
+
+import py_trees
+import carla
+
+
+from srunner.scenariomanager.carla_data_provider import CarlaDataProvider
+from srunner.scenariomanager.scenarioatomics.atomic_behaviors import SwitchOutsideRouteLanesTest, ActorSource, ActorTransformSetter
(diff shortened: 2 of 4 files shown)


srunner/scenarios/parking_exit.py
```diff
@@ -0,0 +1,174 @@
...
+    def __init__(self, world, ego_vehicles, config, debug_mode=False, criteria_enable=True,
+                 timeout=180):
+        """
+        Setup all relevant parameters and create scenario
+        and instantiate scenario manager
+        """
+        self._world = world
+        self._map = CarlaDataProvider.get_map()
+        self.timeout = timeout
+        self._parking_point = convert_dict_to_location(
```
glopezdiest, line 79: Not sure this is needed. We can automatise this by getting the side lane. Or add a parameter like "left" or "right" instead, which is much user friendly than a hardcoded position

srunner/scenarios/parking_exit.py
```diff
@@ -0,0 +1,174 @@
...
+        else:
+            self._front_vehicle_distance = 20  # m
+
+        if 'behind_vehicle_distance' in config.other_parameters:
+            self._behind_vehicle_distance = float(
+                config.other_parameters['behind_vehicle_distance']['value'])
+        else:
+            self._behind_vehicle_distance = 5  # m
+
+        if 'end_distance' in config.other_parameters:
```
glopezdiest, line 94: Again, not really needed, easily automatizable. We have the distance of the vehicle in front, so just use that distance plus some value

srunner/scenarios/parking_exit.py
```diff
@@ -0,0 +1,174 @@
...
+        parking_transform.location = self._parking_point
+        # To get rotation like this because the return value of the
+        # get_waypoint(self, location, project_to_road=True, lane_type=carla.LaneType.Parking)
+        # would be weird in some maps. So using get_waypoint(self._parking_point) to get a waypoint in driving lane
+        # which has the same rotation with the parking point is a safe choice.
+        parking_transform.rotation = self._map.get_waypoint(
+            self._parking_point).transform.rotation
+
+        # Put blocking vehicles
+        front_transform = get_same_dir_transform_by_distance(
```
glopezdiest, line 125: Use the waypoint.next() o waypoint.previous(). The yaw will easily fail  if the road is not completely straight (in both elevation and turns)

srunner/scenarios/parking_exit.py
```diff
@@ -0,0 +1,174 @@
...
+            parking_transform, self._front_vehicle_distance)
+        front_transform.location.z += 0.5
+        behind_transform = get_same_dir_transform_by_distance(
+            parking_transform, -1*self._behind_vehicle_distance)
+        behind_transform.location.z += 0.5
+
+        actor_source_front = ActorSource(
```
glopezdiest, line 132: The actor creation can be done when initializing the scenario. In that way, we can use the CarlaDataProvider.request_new_actor(), instead of an atomic behavior.

Also, ActorSource isn't really meant to be used like this

srunner/scenarios/parking_exit.py
```diff
@@ -0,0 +1,174 @@
...
+        # would be weird in some maps. So using get_waypoint(self._parking_point) to get a waypoint in driving lane
+        # which has the same rotation with the parking point is a safe choice.
+        parking_transform.rotation = self._map.get_waypoint(
+            self._parking_point).transform.rotation
+
+        # Put blocking vehicles
+        front_transform = get_same_dir_transform_by_distance(
+            parking_transform, self._front_vehicle_distance)
+        front_transform.location.z += 0.5
+        behind_transform = get_same_dir_transform_by_distance(
```
glopezdiest, line 128: Same, use the waypoints instead

srunner/scenarios/parking_exit.py
```diff
@@ -0,0 +1,174 @@
...
+            parking_transform, -1*self._behind_vehicle_distance)
+        behind_transform.location.z += 0.5
+
+        actor_source_front = ActorSource(
+            ['vehicle.audi.tt', 'vehicle.tesla.model3', 'vehicle.nissan.micra'],
+            front_transform, 1, "ParkingExit_front_transform_queue", 1)
+
+        actor_source_behind = ActorSource(
```
glopezdiest, line 136: Same

srunner/scenarios/parking_exit.py
```diff
@@ -0,0 +1,174 @@
...
+        sequence.add_child(SwitchOutsideRouteLanesTest(True))
+
+        return sequence
+
+    def _create_test_criteria(self):
+        """
+        A list of all test criteria will be created that is later used
+        in parallel behavior tree.
+        """
+        return [CollisionTest(self.ego_vehicles[0])]
```
glopezdiest, line 167: I forgot to mention this in the previous scenario. For routes, the criteria should be empty, as the collisiontest is already part of the whole route. Add this 

```
if self.route_mode:
    return[]
```

srunner/scenarios/parking_exit.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def __del__(self):
+        """
+        Remove all actors and traffic lights upon deletion
+        """
+        print("del scenario")
```
glopezdiest, line 173: Remove

srunner/scenariomanager/scenarioatomics/atomic_criteria.py
```diff
@@ -1020,9 +1020,15 @@ def update(self):
         if location is None:
             return new_status
 
-        # Check if outside route lanes
-        self._is_outside_driving_lanes(location)
-        self._is_at_wrong_lane(location)
+        # Deactivate by other components
+        activate = py_trees.blackboard.Blackboard().get('AC_SwitchOutsideRouteLanesTest')
```
glopezdiest, line 1024: Do something like the background manager, where you get the blackboard variable, adn then you set it to None

srunner/scenariomanager/scenarioatomics/atomic_criteria.py
```diff
@@ -1020,9 +1020,15 @@ def update(self):
...
+        if activate is not None and activate is False:
```
glopezdiest, line 1025: I'd recommend against this. Better to just do the check [here](https://github.com/carla-simulator/scenario_runner/blob/leaderboard-2.0/srunner/scenariomanager/scenarioatomics/atomic_criteria.py#L1051)

glopezdiest (changes requested):

srunner/scenarios/parking_exit.py
TayYim: Added parking_lane_side=left/right parameter

srunner/scenarios/parking_exit.py
TayYim: end_distance parameter removed. Can be calculated by front_vehicle_distance

TayYim: end_distance parameter removed. Can be calculated by front_vehicle_distance

srunner/scenarios/parking_exit.py
TayYim: Yep. That's better. Implemented.

srunner/scenarios/parking_exit.py
TayYim: Using CarlaDataProvider.request_new_actor() instead of ActorSource

srunner/scenarios/parking_exit.py
TayYim: done

srunner/scenarios/parking_exit.py
TayYim: done

srunner/scenarios/parking_exit.py
TayYim: Added

srunner/scenarios/parking_exit.py
TayYim: Removed

srunner/scenariomanager/scenarioatomics/atomic_criteria.py
TayYim: Done

srunner/scenariomanager/scenarioatomics/atomic_criteria.py
TayYim: Moved this part of code to https://github.com/carla-simulator/scenario_runner/blob/leaderboard-2.0/srunner/scenariomanager/scenarioatomics/atomic_criteria.py#L1051

srunner/scenarios/parking_exit.py
```diff
@@ -0,0 +1,153 @@
...
+        if self._parking_lane_side == "left":
+            parking_waypoint = self._map.get_waypoint(
+                self.ego_vehicles[0].get_location()).get_left_lane()
+        else:
+            parking_waypoint = self._map.get_waypoint(
+                self.ego_vehicles[0].get_location()).get_right_lane()
+
+        front_points = parking_waypoint.next(self._front_vehicle_distance)
+        if len(front_points) > 0:
+            CarlaDataProvider.request_new_actor(
```
glopezdiest, line 110: Better to add a rolename to the vehicle plus some variability to the blueprint with

`actor = CarlaDataProvider.request_new_actor('vehicle.*', transform, rolename='scenario', attribute_filter={'base_type': 'car'})`

Additionally, if the actors will be spawned at the start, it is better to do so at the `initialise_actors` function. [Here](https://github.com/carla-simulator/scenario_runner/blob/master/srunner/scenarios/object_crash_vehicle.py#L62) is an example

srunner/scenarios/parking_exit.py
```diff
@@ -0,0 +1,153 @@
...
+                'vehicle.tesla.model3', front_points[0].transform)
+
+        behind_points = parking_waypoint.previous(
+            self._behind_vehicle_distance)
+        if len(behind_points) > 0:
```
glopezdiest, line 115: Same as before, both the if statement and the actor spawning

srunner/scenarios/parking_exit.py
glopezdiest, line 110: It is also common practice to add the to the `self.other_actors` list, to keep track of all spawned vehicles, in case something fails

srunner/scenarios/parking_exit.py
```diff
@@ -0,0 +1,153 @@
...
+            policy=py_trees.common.ParallelPolicy.SUCCESS_ON_ONE)
+
+        end_condition = py_trees.composites.Sequence()
+        end_condition.add_child(InTriggerDistanceToLocation(
+            self.ego_vehicles[0], parking_waypoint.transform.location, self._end_distance, operator.gt, name="EndTrigger"))
+        root.add_child(end_condition)
+        sequence.add_child(root)
+
+        sequence.add_child(SwitchOutsideRouteLanesTest(True))
+
```
glopezdiest, line 136: Just a comment here, there is no behavior that destroys the actor after the scenario, so they'll stay for the rest of the simulation. If this is intended (it doesn't matter either way), then so be it

srunner/scenariomanager/scenarioatomics/atomic_criteria.py
```diff
@@ -1049,6 +1042,20 @@ def update(self):
 
                 self._current_index = index
 
+        # Deactivate/Activate cheching by blackboard message
+        activate = py_trees.blackboard.Blackboard().get('AC_SwitchOutsideRouteLanesTest')
```
glopezdiest, line 1046: NO! What I was asking for is to change [this](https://github.com/carla-simulator/scenario_runner/blob/master/srunner/scenariomanager/scenarioatomics/atomic_criteria.py#L1125) conditional to also take into account that blackboard. To something like:

```py
if self._active and (self._outside_lane_active or self._wrong_lane_active):
      self._wrong_distance += new_dist
```

Additionally:

```py
active = py_trees.blackboard.Blackboard().get('AC_SwitchOutsideRouteLanesTest')
if active is not None:
    self._active = active
    py_trees.blackboard.Blackboard().set("AC_SwitchOutsideRouteLanesTest", None, overwrite=True)

self._is_outside_driving_lanes(location)
self._is_at_wrong_lane(location)

if self._active and (self._outside_lane_active or self._wrong_lane_active):
      self.test_status = "FAILURE"
```

srunner/scenarios/parking_exit.py
```diff
@@ -0,0 +1,153 @@
...
+        # Teleport ego to the parking point
+        sequence.add_child(ActorTransformSetter(
+            self.ego_vehicles[0], parking_waypoint.transform))
+
+        root = py_trees.composites.Parallel(
+            policy=py_trees.common.ParallelPolicy.SUCCESS_ON_ONE)
+
+        end_condition = py_trees.composites.Sequence()
+        end_condition.add_child(InTriggerDistanceToLocation(
+            self.ego_vehicles[0], parking_waypoint.transform.location, self._end_distance, operator.gt, name="EndTrigger"))
```
glopezdiest, line 131: Remove the `operator.gt` to avoid having to import `operator`

glopezdiest (changes requested):

srunner/scenariomanager/scenarioatomics/atomic_criteria.py
TayYim: Get it. Done.

srunner/scenarios/parking_exit.py
```diff
@@ -0,0 +1,153 @@
...
+        # Put blocking vehicles
+        if self._parking_lane_side == "left":
+            parking_waypoint = self._map.get_waypoint(
+                self.ego_vehicles[0].get_location()).get_left_lane()
+        else:
+            parking_waypoint = self._map.get_waypoint(
+                self.ego_vehicles[0].get_location()).get_right_lane()
+
+        front_points = parking_waypoint.next(self._front_vehicle_distance)
+        if len(front_points) > 0:
```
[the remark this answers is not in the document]
TayYim: Sure, that will be better. add raise Exception for both situations.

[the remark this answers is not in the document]
TayYim: Done

srunner/scenarios/parking_exit.py
TayYim: Get. Put them into _initialize_actors() and add the actors to self.other_actors

srunner/scenarios/parking_exit.py
TayYim: Done

srunner/scenarios/parking_exit.py
TayYim: Added

srunner/scenarios/parking_exit.py
TayYim: I think there should be less side effect, so I add ActorDestroy to remove them.

srunner/scenarios/parking_exit.py
TayYim: The default value of comparison_operator in InTriggerDistanceToLocation  is operator.lt, here I need to judge whether the distance is greater than the given value, so operator.gt is needed I think.

glopezdiest: Ah, I see. Then don't use that condition. Use the `DriveDistance`, which does a very similar job. You can a bit more distance to the condition, in case the ego doesn't go straight, and that will be good enough

TayYim: Using DriveDistance, it works well. Set a longer _end_distance.

glopezdiest (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/882)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @glopezdiest)

glopezdiest (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/882)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @glopezdiest)

glopezdiest (approved):
> Reviewed 2 of 4 files at r2, 1 of 1 files at r4, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/882)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @TayYim)

srunner/scenarios/parking_exit.py
```diff
@@ -0,0 +1,173 @@
...
+                                          debug_mode,
+                                          criteria_enable=criteria_enable)
+
+    def _initialize_actors(self, config):
+        """
+        Custom initialization
+        """
+        # Put blocking vehicles
+        if self._parking_lane_side == "left":
+            self._parking_waypoint = self._map.get_waypoint(
```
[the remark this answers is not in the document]
TayYim: Understand. That's indeed a more robust way.

glopezdiest (approved):
> Reviewed 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/882)* status: all files reviewed, 1 unresolved discussion (waiting on @glopezdiest)

glopezdiest (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/882)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @TayYim)

PR #886, opened 2022-04-29
add SwitchLane, remove RemoveLane

TayYim:
#### Description

I find that the implementation of RemoveLane is a little rude. It just deletes a lane key from self._road_dict, so it’s hard to recover the lane and keep sourcing BA actors like before.

So I implement the SwitchLane(lane_id, active) for BA. 

If active is False, it will destroy all actors on the lane and deactivate  the source instead of deleting the lane key. 
If active is True, it will make the source active and reset the source waypoint. Because the source waypoint is still on the old location before the sourcing is deactivated, which may be too far from ego’s location currently.

Based on this, I add a recover step to HighwayExit scenario. It works well.

Fixes #  

Replace RemoveLane with SwitchLane

  * **Python version(s):** 3.7

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/886)


srunner/scenarios/background_activity.py
```diff
@@ -1073,23 +1073,24 @@ def _update_road_sources(self, prev_ego_index):
...
-                # Stop the source from being too close to the ego or last lane vehicle
-                actor_dist = max(0, last_location.distance(source_location) - self._spawn_dist)
-                ego_dist = max(0, ego_location.distance(source_location) - min_distance)
-                move_dist = min(actor_dist, ego_dist)
-
-                # Move the source forward if needed
-                if move_dist > 0:
-                    new_source_wps = source.wp.next(added_dist)
-                    if not new_source_wps:
+                if source.actors:
```
glopezdiest, line 1076: I don't like this. Due to stopping this, we have to later on move it back to its correct position. I suggest just changing the next code such that if no actors are found, the source keeps moving at a specific distance from the ego.

srunner/scenarios/background_activity.py
```diff
@@ -1816,16 +1818,25 @@ def _leave_space_in_front(self, space):
...
         lane_id = str(lane_id)
         lane_id_list = [x.split('*')[-1] for x in  list(self._road_dict.keys())]
         if lane_id in lane_id_list:
             lane_index = lane_id_list.index(lane_id)
             lane_key = list(self._road_dict.keys())[lane_index]
-            for actor in list(self._road_dict[lane_key].actors):
-                self._destroy_actor(actor)
-            self._road_dict.pop(lane_key)
+            self._road_dict[lane_key].active = active
+            if active:
```
glopezdiest, line 1832: This the previous change, nothing should be done when activating the lane source.

glopezdiest (changes requested):

srunner/scenarios/background_activity.py
TayYim: Changed: If there are no actors, set the last_location to the ego's location, so that the source waypoint will keep moving forward.

srunner/scenarios/background_activity.py
TayYim: removed

TayYim:
Conflicts resolved

glopezdiest (approved):
> Reviewed 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/886)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @TayYim)

PR #908, opened 2022-06-28
Invading turn

TayYim:
#### Description

1. Add InvadingTurn scenario
2. Add opt_dict to BasicAgentBehavior

Fixes #  

  * **Python version(s):** 3.7

#### Possible Drawbacks

As discussed in discord, when the scenario is triggered, the simulation will be stuck for a second. 
Since the CarlaDataProvider.get_global_route_planner()  is not merged into the leaderboard-2.0 branch yet, this problem may be solved later.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/908)

### DIFF
--- a/Docs/CHANGELOG.md
+++ b/Docs/CHANGELOG.md
@@ -28,6 +28,7 @@
     - ParkingCutIn: A vehicle parked at the side cuts in front of the ego.
     - BlockedIntersection: With low visibility, the ego performs a turn only to find out that the end is blocked by another vehicle.
     - MergerIntoSlowTraffic: the ego has to enter a lane that is filled with slow traffic.
+    - InvadingTurn: the ego is about to turn right when a vehicle coming from the opposite lane invades the ego's lane, forcing the ego to move right to avoid a possible collision.
 * Added new functions to the BackgroundManager
 * Minor improvements to some example scenarios. These include FollowLeadingVehicle, VehicleTurning, DynamicObjectCrossing and SignalizedJunctionRightTurn and RunningRedLight. Their behaviors are now more smooth, robust and some outdated mechanics have been removed
 * SignalizedJunctionLeftTurn has been remade. It now has an actor flow on which the ego has to merge into, instead of a single vehicle.
--- a/srunner/routes/leaderboard_2.0_testing.xml
+++ b/srunner/routes/leaderboard_2.0_testing.xml
@@ -168,17 +168,17 @@
          </scenario>
       </scenarios>
    </route>
-   <route id="68" town="Town02"> <!-- InvadingTurn, old towns -->
+   <route id="68" town="TrainingMap"> <!-- InvadingTurn, new town -->
       <waypoints>
-         <position x="150" y="109.5" z="0.0" />
-         <position x="182" y="109.5" z="0.0" />
-         <position x="189.7" y="121" z="0.0" />
-         <position x="189.7" y="171" z="0.0" />
+         <position x="-1299" y="3639.2" z="368" />
+         <position x="-1283.9" y="3672" z="368" />
+         <position x="-1287.9" y="3681.8" z="368" />
+         <position x="-1440" y="3680" z="368" />
       </waypoints>
       <scenarios>
          <scenario name="InvadingTurn_1" type="InvadingTurn">
-            <trigger_point x="170" y="109.5" z="0.0" yaw="0"/>
-            <adversary_start x="194" y="123" z="0"/>
+            <trigger_point x="-1292.09" y="3657.82" z="365" yaw="63.4"/>
+            <adversary_start x="-1287.3" y="3684.4" z="368"/>
          </scenario>
       </scenarios>
    </route>
--- a/srunner/routes/leaderboard_2.0_testing.xml
+++ b/srunner/routes/leaderboard_2.0_testing.xml
@@ -209,7 +209,7 @@
             <scenario name="InvadingTurn_1" type="InvadingTurn">
                <trigger_point x="1976.1" y="4260.5" z="361" yaw="99"/>
                <distance value="120" />
-               <offset value="-1.1"/>
+               <offset value="0.5"/>
             </scenario>
          </scenarios>
    </route>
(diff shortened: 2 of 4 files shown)


glopezdiest (changes requested):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/908)* status: 0 of 4 files reviewed, 5 unresolved discussions (waiting on @glopezdiest and @TayYim)

___
*[`srunner/scenarios/invading_turn.py` line 37 at r4](https://reviewable.io/reviews/carla-simulator/scenario_runner/908#-N5dHwDmA6EPWClTfq0z:-N5dHwDmA6EPWClTfq1-:b-jys4bj) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/1281074ff05ca085732bd393637702294117c253/srunner/scenarios/invading_turn.py#L37)):*
> ```Python
> 
> 
> class InvadingTurn(BasicScenario):
> ```

Over the iterations, this isn't an invading turn anymore right? We can place this scenario in any road we want, not necesarrily a turn



_Code quote:_
```Python
class InvadingTurn(BasicScenario):
```
___
*[`srunner/scenarios/invading_turn.py` line 59 at r4](https://reviewable.io/reviews/carla-simulator/scenario_runner/908#-N5dIaoeDe7WTCWXIVrA:-N5dIaoeDe7WTCWXIVrB:bdc6hzv) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/1281074ff05ca085732bd393637702294117c253/srunner/scenarios/invading_turn.py#L59)):*
> ```Python
>             self._trigger_location)
> 
>         self._adversary_start = convert_dict_to_location(
> ```

I don't like this parameter. For me, the more natural way would be to add something like `distance`, and the scenario will calcuate the position using the waypoint API



_Code quote:_
```Python
        self._adversary_start = convert_dict_to_location(
            config.other_parameters['adversary_start'])
```
___
*[`srunner/scenarios/invading_turn.py` line 65 at r4](https://reviewable.io/reviews/carla-simulator/scenario_runner/908#-N5dJ1lHAh8I0Paa18Za:-N5dJ1lHAh8I0Paa18Zb:b-w3znmr) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/1281074ff05ca085732bd393637702294117c253/srunner/scenarios/invading_turn.py#L65)):*
> ```Python
>         # How much the adversary will invade the lane. Should be negative.
>         # The greater the absolute value of this parameter, the higher the offset of the vehicle.
>         if 'offset' in config.other_parameters:
> ```

Have you tried having the offset as a percentage of the lane width (i.e, offset of 1 means the adversary is going through the middle of the lane)?
Or like being the width of the vehicle invading the opposite lane (i.e, an offset of 1 means that the opposite vehicle is invading the opposite lane by 1m)?

This type of offset is also correct though, just some brainstorming ideas



_Code quote:_
```Python
        if 'offset' in config.other_parameters:
            self._offset = float(
                config.other_parameters['offset']['value'])
        else:
            self._offset = -1.1
```
___
*[`srunner/scenarios/invading_turn.py` line 94 at r4](https://reviewable.io/reviews/carla-simulator/scenario_runner/908#-N5dIUDi23Kdz4hkareW:-N5dIUDi23Kdz4hkareX:buo40ur) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/1281074ff05ca085732bd393637702294117c253/srunner/scenarios/invading_turn.py#L94)):*
> ```Python
> 
>         spawn_transform.location.z += 5  # Avoid colliding with BA actors
>         actor = CarlaDataProvider.request_new_actor(
> ```

Add the need for vehicles to have lights, like the BA:

https://github.com/carla-simulator/scenario_runner/blob/leaderboard-2.0/srunner/scenarios/background_activity.py#L188



_Code quote:_
```Python
        actor = CarlaDataProvider.request_new_actor(
            "vehicle.*", spawn_transform, rolename='scenario')
```
___
*[`srunner/scenarios/invading_turn.py` line 117 at r4](https://reviewable.io/reviews/carla-simulator/scenario_runner/908#-N5dJixS0bKAUhGDF-HC:-N5dJixS0bKAUhGDF-HD:bljp5r7) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/1281074ff05ca085732bd393637702294117c253/srunner/scenarios/invading_turn.py#L117)):*
> ```Python
> 
>         if self.route_mode:
>             sequence.add_child(SwitchLane(
> ```

I'd prefer using *LeaveSpaceInFront*, which moves the front vehicles forward, resulting in the same effect, BUT without destroying them, which is a currently problem after this scenario ends



_Code quote:_
```Python
            sequence.add_child(SwitchLane(
                self._reference_waypoint.lane_id, False))
```

glopezdiest (commented):
> Reviewed 2 of 4 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/908)* status: all files reviewed, 1 unresolved discussion (waiting on @glopezdiest)

glopezdiest (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/908)* status: all files reviewed, 1 unresolved discussion (waiting on @TayYim)

___
*[`srunner/scenarios/invading_turn.py` line 37 at r4](https://reviewable.io/reviews/carla-simulator/scenario_runner/908#-N5dHwDmA6EPWClTfq0z:-N5nJn8Y3zRrLk0-LNmd:b-3xd1b8) ([raw file](https://github.com/carla-simulator/scenario_runner/blob/1281074ff05ca085732bd393637702294117c253/srunner/scenarios/invading_turn.py#L37)):*
<details><summary><i>Previously, TayYim (Tay Yim) wrote…</i></summary><blockquote>

That's right. Shall we change the name to 'InvadingLane'?
Besides, I think, if it happens in a turn, it's more challenging to detect the adversary in advance than if it happens on a straight road. So if we need to keep the original idea, a turn is better.
</blockquote></details>

Yeah, I guess the idea of doing it at a turn is still there for visibility purposes. Okay, let's leave it like that then

glopezdiest (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/908)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @TayYim)

PR #946, opened 2022-11-01
Update atomic_trigger_conditions.py

TigerStone93:
#### Description

In the `StandStill` class, change the `EPSILON`, `0.001`, to the float value `0.1`.
When a vehicle stops after cruising, even though it actually not moves at all, but the `velocity` could have a value larger than `0.001` while the displayed velocity is 0kph, so the `StandStillCondition` is not triggered as the developer's original intention. 
This change makes `StandStill` class work properly not only before vehicle's start but also after the vehicle's stop.

  * **Python version(s):** 3.7
  * **CARLA version:** 0.9.13

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/946)


glopezdiest (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/946)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @TigerStone93)

PR #977, opened 2023-03-28
Adding a setup.py file.

axelbr:
#### Description

We added a setup.py file, which makes it possible to directly install scenario runner with a single command (e.g. pip install).

  * **Python version(s):** 3.8
  * **Unreal Engine version(s):** n.a.
  * **CARLA version:** 0.9.13

#### Possible Drawbacks

We might have missed some artifacts in setup.py. Please check.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/977)


WaelKarkoub (commented):
Will you be willing to split this PR up into to? I was working on just adding the `setup.py` but I noticed you already did the work

PR #1091, opened 2024-06-26
Fix dead actor accumulation in CDP

Daraan:
I realized that in the `CarlaDataProvider._carla_actor_pool` more and more dead actors accumulated over time when scenarios run. The reason is that actors are created over the CDP and are added to the pool, however, they are not destroyed over the CDP.  
I think I located most of the locations where this is the case - however I think there is still at least one missing.

As a side result it the `remove_actor_by_id` function is now also slightly more informative. If the return value is `None` the users needs to take care of the destruction.

---

I also propose an update to the `CarlaDataProvider._actor_[velocity, location, transform]_map`, while this new code is in itself more expensive it was in the long run still faster as the sizes were overall smaller. Of course, just talking about a few *ms*.

---

### Tested on 

    Platform(s): Ubuntu
    Python version(s): 3.7 and 3.10
    Unreal Engine version(s): 4.24
    CARLA version: 0.9.15 dev

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/1091)

### DIFF
--- a/srunner/scenariomanager/carla_data_provider.py
+++ b/srunner/scenariomanager/carla_data_provider.py
@@ -1003,11 +1003,12 @@ class CarlaDataProvider(object):  # pylint: disable=too-many-public-methods
         Remove an actor from the pool using its ID
         """
         if actor_id in CarlaDataProvider._carla_actor_pool:
-            CarlaDataProvider._carla_actor_pool[actor_id].destroy()
-            CarlaDataProvider._carla_actor_pool[actor_id] = None # type: ignore
+            was_destroyed = CarlaDataProvider._carla_actor_pool[actor_id].destroy()
+            CarlaDataProvider._carla_actor_pool[actor_id] = None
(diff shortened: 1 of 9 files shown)


glopezdiest (approved):
> Reviewed 9 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/1091)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Daraan)

glopezdiest:
I'll check this in more detail over the next week to make sure of all the details but this is a great PR, thanks you very much.

Daraan:
I am confident that it *should* work. If the actor is registered, which all should be, they will be destroyed by `                CarlaDataProvider.remove_actor_by_id(actor.id)`. If one isn't, or was already removed, you will get a warning print and `None` is returned instead of `False.`

PR #1095, opened 2024-07-03
 [Help wanted] Update requirements.txt for Python > 3.7

Daraan:
The current requirements are quite restrictive and outdated for newer python versions. I think we should update the requirements file with some version checks to allow Python3.7+ and the latest supported 3.10 to install and work smoothly

- opencv_python & numpy are the most important.
- networkx breaks in Python 3.10; needs at least networkx 2.7 if I recall correctly.
- Shapely needs at last +1 minor version
- ...

**Maybe you can share your installations from 3.8 - 3.10 and help on this draft to craft a working version for the newer python versions.**

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/1095)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,12 +1,29 @@
+# Python 3.7 - 3.10
+opencv-python==4.2.0.32; python_version <= "3.7"
+opencv-python~=4.5.5.64,<=4.5.5.64; python_version < "3.10"
+opencv-python==4.5.5.64; python_version == "3.10"
+
+# opencv 4.5+ will install numpy>=1.21.2
+numpy==1.18.4; python_version >= '3.0' and python_version <= "3.7"
+
+# 2.2 breaks in 3.10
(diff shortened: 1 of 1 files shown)


GVMNREDDY:
is there any problem to work on i will be available to do

Daraan:
> is there any problem to work on i will be available to do

Thank you for your reply. The idea is to get an requirements.txt file that works with multiple python versions. Some requirements like `opencv-python`, `numpy` are troublesome to figure out to get a valid combination for all python versions.

I only work with python 3.7 & 3.10 and no capacity to check the other versions in between.

So if you are willing try to to get the scenario_runner running with > 3.7 and share your requirements of the important packages :)

Daraan:
Closing this for now as requirements were updated, for <= python 3.10 versions it might still be a useful reference.

PR #1096, opened 2024-07-03
Fix: Remove duplicate clean method calls

starrye:
#### Description

Fixes #  

  * **Platform(s):** Ubuntu
  * **Python version(s):** 3.8
  * **Unreal Engine version(s):** 5
  * **CARLA version:** 0.9.15

#### Possible Drawbacks

Hello, when I added custom logic to CarlaDataProvider clean, I found that it was called many times. And I saw in Class `ScenarioRunner _cleanup` method calls the `self.manager. cleanup ()` and `CarlaDataProvider. cleanup ()`, Of the `self.manager.cleanup ()` includes `CarlaDataProvider.cleanup()`, so can make `CarlaDataProvider.cleanup()` method is called three times, In fact, after stopping the operation in `ScenarioManager.run_scenario()`, `self.cleanup()`will be called immediately, which already contains`CarlaDataProvider.cleanup()`, So consider removing the redundant clean method call in the _cleanup method of the ScenarioRunner class

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/1096)


starrye:
@PabloVD please review

starrye:
@PabloVD new commit please review Fix Atomic_behaviors AddActor Func

PabloVD:
Looks ok to me, merged, thanks!

PR #1098, opened 2024-07-08
Speed up information retrieval

Daraan:

  * **Python version(s):** Python 3.7 and 3.10
  * **CARLA version:** 0.9.15

---

When using `CarlaDataProvider.get_velocity / transform / location(actor)` the lookup function loops over all dict keys and compares the actor id of each key against the argument's id. Comparing all ids is necessary, as different python instances of the same actor have different hashes, but inefficient if the argument is a present key.

This PR adds a best-case lookup before and checks if the actor a present key before it falls back to the slower loop over all actors.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/1098)


glopezdiest (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/1098)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Daraan)

PR #1110, opened 2024-08-27
`GameTime.get_wallclocktime()` should return a `datetime` object

Daraan:
#### Description

When using `self.wallclock_t0 = GameTime.get_wallclocktime()`, which returns  `_platform_timestamp` a type-checker complains that the type used afterwards is `int | datetime.datetime`, however it is never used as `int` this is just the intial value. This minimal PR unifies the type to be always a `datetime` object.

```diff
- _platform_timestamp = 0
+ _platform_timestamp = datetime.datetime.fromtimestamp(0)
```

  * **Python version(s):** 3.7 & 3.10
  * **Unreal Engine version(s):** 
  * **CARLA version:** 0.9.15 dev

#### Possible Drawbacks

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/1110)


Daraan:
Hey @glopezdiest can we add this. For pyright (and other type checkers that implicitly pick up types) users it will reduce some warnings.

PR #1113, opened 2024-09-11
`spawn_point` for CarlaDataProvider.request_new_actor should be optional when using `random_location`. 

Daraan:
The `request_new_actor` of the CarlaDataProvider has arguments for `spawn_point` and for `random_location`.

`spawn_point` is a **required argument, however it is unused if `random_location` is True**

This PR makes `spawn_point` an optional argument and ensures that at least one of these two is provided and warns if both are provided. **Behavior of current code will not change.**

### DIFF
--- a/srunner/scenariomanager/carla_data_provider.py
+++ b/srunner/scenariomanager/carla_data_provider.py
@@ -759,12 +759,16 @@ class CarlaDataProvider(object):  # pylint: disable=too-many-public-methods
         return actor
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


glopezdiest (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/scenario_runner/1113)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Daraan)

glopezdiest:
Makes perfect sense, approved

